Issue #17, opened 2017-10-19
Error: Evaluation failed: Error: blob:http://localhost:45678/f778a6eb-f084-4f21-9495-c03388a074a5 not in stylesheetAstObjects!


stereobooster:
Your project is very helpful. My current pages prerendered with [react-snap](https://github.com/stereobooster/react-snap) start to render at 0.5 seconds and have about 80% of content painted. But main js bundle is 200+kb (because I use [react-mapbox-gl](https://github.com/alex3165/react-mapbox-gl), which is pretty heavy). So I decided to use dynamic imports and [react-loadable](https://github.com/thejameskyle/react-loadable). But now I get this error.

```
events.js:182
      throw er; // Unhandled 'error' event
      ^

Error: Evaluation failed: Error: blob:http://localhost:45678/f778a6eb-f084-4f21-9495-c03388a074a5 not in stylesheetAstObjects!
    at links.filter.forEach.stylesheet (<anonymous>:125:19)
    at Array.forEach (<anonymous>)
    at stylesheetAstObjects (<anonymous>:123:10)
    at <anonymous>:159:7
    at Frame._rawEvaluate (/Users/slavik/my/map-react/node_modules/minimalcss/node_modules/puppeteer/lib/FrameManager.js:250:13)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
error Command failed with exit code 1.
```

UPD: this happens because of [webworkify](https://github.com/browserify/webworkify/blob/master/index.js), which is used by mapbox-gl.

```js
var blob = new Blob(['alert(1)'], { type: 'text/javascript' });
URL.createObjectURL(blob);
"blob:https://github.com/47b9e2f7-f81f-49a0-a14b-5a1302976986"
```

this is how to read those files
```js
fetch("blob:http://127.0.0.1:8080/7047145b-2f81-4b95-901b-66e4dca073ae").then( async (x) => console.log(await x.text()) )
```

peterbe:
I'm not familiar with this webworkify thing. Also, I don't see how it connects to react-loadable. 

Anyway, is this a trick to load more JavaScript later? Or if I understand (unlikely) what react-loadable is, it's a wrapper for the webpack tricks that can load more JS later (and in the case of react-loadable, it's done based on certain components being needed). 

Either way. The goal has to be for minimalcss to stay away from how loading works as much as possible and simply wait till the network is idle and then focus on downloaded CSS and the DOM. 

[...]

PR #18, opened 2017-10-19
Handle blob: urls

stereobooster:
Fix for https://github.com/peterbe/minimalcss/issues/17 There is example file for test


src/run.js
```diff
@@ -191,9 +190,20 @@ const minimalcss = async options => {
...
+              const response = await fetch(stylesheet.href)
+              const text = await response.text()
+              const ast = csstree.parse(text, {
+                parseValue: false,
+                parseRulePrelude: false
+              })
+              stylesheetAstObjects[url] = csstree.toPlainObject(ast)
+              stylesheetContents[url] = text
+            } else {
+              throw new Error(`${stylesheet.href} not in stylesheetAstObjects!`)
```
stereobooster: much simpler solution would be ignore `blob:` urls

src/run.js
```diff
@@ -191,9 +190,20 @@ const minimalcss = async options => {
...
           )
         })
-        .forEach(stylesheet => {
+        .map(async stylesheet => {
           if (!stylesheetAstObjects[stylesheet.href]) {
-            throw new Error(`${stylesheet.href} not in stylesheetAstObjects!`)
+            if (stylesheet.href.startsWith('blob:')) {
+              const response = await fetch(stylesheet.href)
+              const text = await response.text()
+              const ast = csstree.parse(text, {
```
stereobooster: `csstree` is not defined in this context :/

src/run.js
```diff
@@ -48,18 +48,17 @@ const minimalcss = async options => {
...
-            parseRulePrelude: false
-          })
-          stylesheetAstObjects[url] = csstree.toPlainObject(ast)
-          stylesheetContents[url] = text
+        const text = await response.text()
+        const ast = csstree.parse(text, {
+          parseValue: false,
+          parseRulePrelude: false
         })
+        stylesheetAstObjects[url] = csstree.toPlainObject(ast)
```
stereobooster: Or

```js
    if (url.startsWith('blob:')) {
          page.evaluate(url => {
            return fetch(url).then(response => response.text())
          }, url).then(text => {
            console.log(text)
            const ast = csstree.parse(text, {
              parseValue: false,
              parseRulePrelude: false
            })
            stylesheetAstObjects[url] = csstree.toPlainObject(ast)
            stylesheetContents[url] = text
          })
        } else {
          response.text().then(text => {
            const ast = csstree.parse(text, {
              parseValue: false,
              parseRulePrelude: false
            })
            stylesheetAstObjects[url] = csstree.toPlainObject(ast)
            stylesheetContents[url] = text
          })
        }
```

src/run.js
```diff
@@ -193,7 +193,12 @@ const minimalcss = async options => {
         })
         .forEach(stylesheet => {
           if (!stylesheetAstObjects[stylesheet.href]) {
-            throw new Error(`${stylesheet.href} not in stylesheetAstObjects!`)
+            if (stylesheet.href.startsWith('blob:')) {
+              // TODO: do we need to do something here?
```
peterbe: I think this is the right thing to do. It's sad but it's what it is. 

It's not that different from the fact that `minimalcss` doesn't clean up any `<style>` tags on the page. (Perhaps it should in a future version. But that's a story for another day) 
If someone uses a blob to inject a bunch of "fat" CSS then I think it's best to just ignore it. 

However, I would like one change. Can you remove this if statement and weave it into the filter loop a couple of lines above?
Perhaps something like this:
```javascript
.filter(link => {
    return (
       link.href && !link.href.startsWith('blob:') &&
       ...
    )
})
```
What do you think about that?

stereobooster:
Done

peterbe:
Great work! I think this deserves a version bump.

peterbe:
0.2.3 out.

Issue #19, opened 2017-10-21
Prevent network request to Google analytics


stereobooster:
In `reac-snap` I prevent GA request by detecting userAgent. But we can block those requests based on domain name - there is a comment in source in how to do this.

peterbe:
I saw your PR and I love that you're contributing! Even if I don't approve of everything it inspires me to work on this lib more. 

The example of userAgent is a good example. I think it's a mistake to prevent network requests on just domain. That might be enough in *your* case. Or *his*. Or *hers*. `minimalcss` already has some strong opinions about URLs that shouldn't be loaded, which is that images are not needed. 

The most flexible solution is to be able to pass in a callable where the user can decide for herself. E.g.

```javascript
const skippable = request => {
   const url = response.url
   const ct = response.headers['content-type'] || ''
   if (ct === 'something-special') {
      return true
   } 
   if (url.indexOf('special.js') > -1) {
      return true
   }
   // assume to not skip it
  return false
}
const options = { 
  urls: ['http://peterbecom.dev/css-blocking/ultra-basic.html'],
  skippable
 }
minimalcss
  .minimize()
  .then(result => {
    ...
```

And in `src/run.js`, somewhere, we'd do this:
```javascript
page.on('request', request => {
    if (options.skippable && options.skippable(request)) {
      request.abort()
    }
    ...
```

stereobooster:
I see in case of `minimalcss` it probably doesn't make sense to maintain a blacklist, because this is a tool. But I want some kind of default solution for `react-snap` because this is a tool which should work out of the box.

Possible solutions:

1. blacklist
2. block every request to other domain by default + whitelist
3. callback function

Obviously, a callback function is the most flexible solution (for a lib, but not for the tool).

peterbe:
One tricky thing is that if you "import" minimalcss it's obvious how to write a callable function. But if you use the minimalcss cli, it's not obvious. You *could* do something like `./bin/minimalcss --skip www.google-analytics.com --skip tracking.ads.com http://www.example.com` and the cli could convert that list of domains into a function "for you".

stereobooster:
My current focus is on using `minimalcss` as a library. 

This doesn't look like elegant solution:

```sh
./bin/minimalcss --skip www.google-analytics.com --skip tracking.ads.com
```

peterbe:
So what's left is the `--skip` options to the cli.

PR #20, opened 2017-10-21
Prevent network request to Google analytics aka Blacklist functionality

stereobooster:
Fix for https://github.com/peterbe/minimalcss/issues/19


src/blacklist.js
```diff
@@ -0,0 +1,12 @@
+const Url = require("url");
```
peterbe: Curious that Prettier didn't catch this. It should be single quotes. 
Perhaps the config for Prettier is wrong.

src/blacklist.js
```diff
@@ -0,0 +1,12 @@
...
+
+const blacklist = {
+  'www.google-analytics.com': true,
```
peterbe: When is the value ever `false`?
Why not make this an array or a set? Sets are as fast to look up as dictionaries.

src/run.js
```diff
@@ -36,13 +37,14 @@ const minimalcss = async options => {
       if (/data:image\//.test(request.url)) {
         // don't need to download those
         request.abort()
-      } else if (/\.(png|jpg|jpeg$)/.test(request.url)) {
+      } else if (/\.(png|jpg|jpeg|webp$)/.test(request.url)) {
         request.abort()
       } else if (stylesheetAstObjects[request.url]) {
         // no point downloading this again
         request.abort()
+      } else if (isInBlacklist(request.url)) {
```
peterbe: This should be...
```javascript
} else if (isInBlacklist(options.blacklist, request.url)) {
```
That way, the "user/client" can control the blacklist. We can certainly maintain a little default list of popular URLs like `www.google-analytics.com`. But if we do maintain a list it should be lifted out and put somewhere else. like in this `src/run.js` file (near the top and call it `BLACKLIST` not `blacklist`)

src/blacklist.js
peterbe: Don't rename the module name as if it was a class. Call it `url` here and rename the argument in the `isInBlacklist` function to `uri` or `url_`.

src/blacklist.js
```diff
@@ -0,0 +1,12 @@
...
+  'api.mapbox.com': true,
+  'a.tiles.mapbox.com': true,
+  'b.tiles.mapbox.com': true,
+}
+
+const isInBlacklist = url => blacklist[Url.parse(url).hostname]
```
peterbe: I don't feel comfortable with this. It feels too "brutal". I don't even know what `api.mapbox.com` is but imagine it's a REST endpoint that gives the web page the JSON it needs to paint the screen in a certain way onload. Then blacklisting it would prevent the DOM becoming what it becomes for normal browser users. 

First of all, let's minimize the list of default domains in the list. Second, I wonder if we should do more to filter than just look at the domain. What if you want *some* assets from a certain domain but skip all others. For example, you might to blacklist `a.tiles.mapbox.com/*.js`.

peterbe: The name isn't ideal. That fact that it's a list internally is not important externally. The function can be called `isBlacklisted` or `blacklistedUrl` or perhaps even more user friendly is `skippableUrl`. What do you think? To me the word "blacklisted" means something very permanent. Like things that are set in stone and unlikely to change other than more potentially being added.

src/run.js
```diff
@@ -36,13 +37,14 @@ const minimalcss = async options => {
       if (/data:image\//.test(request.url)) {
         // don't need to download those
         request.abort()
-      } else if (/\.(png|jpg|jpeg$)/.test(request.url)) {
+      } else if (/\.(png|jpg|jpeg|webp$)/.test(request.url)) {
```
peterbe: We probably want to skip `.gif` too. And I bet there are a bunch others we've forgotten too.

src/blacklist.js
stereobooster: Ah yes, I forgot about new ES features. Set is what I was looking for

src/run.js
stereobooster: well let's start with something and will see. I will add gif

stereobooster:
How about this version?

stereobooster:
What if CLI will skip all third-party requests by default (except css obviously)?

src/blacklist.js
stereobooster: 1. Yes agree `api.mapbox.com` is too specific. 
2. `a.tiles.mapbox.com` provides tiles which are either images or binary vector files (protobuf file format). It would make sense to blacklist `*.tiles.mapbox.com`

src/run.js
```diff
@@ -36,13 +36,14 @@ const minimalcss = async options => {
       if (/data:image\//.test(request.url)) {
         // don't need to download those
         request.abort()
-      } else if (/\.(png|jpg|jpeg$)/.test(request.url)) {
+      } else if (/\.(png|jpg|jpeg|webp|gif$)/.test(request.url)) {
```
peterbe: There's a bug in this line. It was there before you made it better :)

https://github.com/peterbe/minimalcss/issues/24

I'm going to try to fix that now. Afterwards, can you rebase up master?

src/run.js
```diff
@@ -36,13 +36,14 @@ const minimalcss = async options => {
...
         request.abort()
       } else if (stylesheetAstObjects[request.url]) {
         // no point downloading this again
         request.abort()
+      } else if (options.skippable && options.skippable(request)) {
```
peterbe: Excellent! I think this is a great start. But it's currently undocumented and that's almost more important than the functionality. 

Can you update the README about this being a possible options key and add an example (perhaps one with your mapbox URLs). 
In fact, under the "API" section of the README you can just mention it's a callable that gets a Request object and expects to return a boolean if defined. 

Then add another section where you describe an example of a more elaborate version. Perhaps an example that uses those mapbox.com URLs. Also, it would be cool if the example demonstrates how you can also skip the URL based on the Content-Type header.

stereobooster: but this is request, not response. I suppose it does not have Content-Type header.

src/run.js
peterbe: If you remove your change and the rebase against master the conflict will go away.

src/run.js
peterbe: You're so totally right. Sorry, I rushed that comment. Ignore that bit about `content-type`.
However, your documentation should point to Request on this URL: https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-request

peterbe:
> What if CLI will skip all third-party requests by default (except css obviously)?

First of all, the solution we're talking about is blocking requests *independent* of it being the CLI or used as a module. 
We already do this to some degree. The assumption is that you don't ever need the images to get a complete DOM. But it makes me nervous actually. What if someone has a piece of code like this:
```javascript
window.onload = function() {
   var img = document.createElement('img');
   img.src = '/image.png';
   img.onload = function() {
       // Only insert this after the image has been loaded
       var caption = document.createElement('div');
       caption.classList.add('caption');
       document.getElementById('#foo').appendChild(caption);
   }
   document.getElementById('#foo').appendChild(img);
};
```
Our default blocking of images would affect how the DOM solidifies when we do the minimal CSS analysis. 

Perhaps the solution to this is an option (also exposed in the cli) where you can optionally load all images.

peterbe:
> Or what if lib will have callback, but CLI will have default blacklist?

I disagree. The CLI should just be about executing the main code from the command line.

stereobooster:
done

stereobooster:
Can you bump this as npm package too? 🙏

peterbe:
Will do. That an the new image loading override option.

Issue #21, opened 2017-10-21
Use some-kind of type validator


stereobooster:
For example puppeteer uses TypeScript to validate code based on jsdoc annotations https://github.com/GoogleChrome/puppeteer/pull/986

Also flow can do some validation, but not sure if there is puppeteer type definitions.

peterbe:
Are you suggesting that we rewrite it in TypeScript? I wouldn't mind.

peterbe:
Worth thinking about the fact the minimalcss isn't much of a lib. It's just an elaborate script around something much more powerful.

stereobooster:
We can use TypeScript for checking types in JS files. 

> Worth thinking about the fact the minimalcss isn't much of a lib

Exactly. It is very hard to write unit tests, only integration tests make sense. But with type checker we can verify that we avoid **some** class of errors without huge test cases.

peterbe:
Sure. I'm up for it. I've never actually used TypeScript and I'm eager to try it. Will you lead the way?

PR #22, opened 2017-10-23
typescript checker

stereobooster:
proof of concept for #21 

usage:

```sh
tsc
```

not useful so far, but lets see how it goes

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,15 @@ key is `urls`. Other optional options are:
 * `debug` - all console logging during page rendering are included in the
   stdout.
 
+## Development
+
+We use ES6+ with jsdoc comments and TypeScript to do static type checking, like [puppeeteer does](https://github.com/GoogleChrome/puppeteer/pull/986/files).
+
+Run `tsc` to check types.
+
+```sh
+yarn run tsc
+```
 
 ## License
 
--- a/package.json
+++ b/package.json
@@ -30,6 +30,8 @@
     "puppeteer": "^0.10.2"
   },
   "devDependencies": {
+    "@types/node": "^8.0.46",
+    "@types/puppeteer": "^0.12.0",
     "husky": "^0.14.3",
     "lint-staged": "^4.2.1",
     "prettier": "^1.7.0"
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "@types/puppeteer": "^0.12.0",
     "husky": "^0.14.3",
     "lint-staged": "^4.2.1",
-    "prettier": "^1.7.0"
+    "prettier": "^1.7.0",
+    "typescript": "^2.7.0-dev.20171025"
   },
   "scripts": {
     "precommit": "lint-staged"
--- a/src/run.js
+++ b/src/run.js
@@ -1,7 +1,9 @@
 'use strict'
 
 const puppeteer = require('puppeteer')
+// @ts-ignore
 const csso = require('csso')
+// @ts-ignore
 const csstree = require('css-tree')
 const collectImportantComments = require('./utils').collectImportantComments
 
--- a/src/run.js
+++ b/src/run.js
@@ -7,6 +7,11 @@ const csso = require('csso')
 const csstree = require('css-tree')
 const collectImportantComments = require('./utils').collectImportantComments
 
+/**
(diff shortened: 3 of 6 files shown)


peterbe:
I'm guessing this requires a global install of `tsc`. Is it not something that can be part of the `devDependencies`?

peterbe:
Cool. It works!
```
dev/JAVASCRIPT/minimalcss  (a10f608...) ✗                                                                                                                                                                     ▶ tsc
src/run.js(197,16): error TS2532: Object is possibly 'undefined'.
src/run.js(200,23): error TS2532: Object is possibly 'undefined'.
```

Now to land this, can you please add a blurb about this in the README?
Hopefully we can expand that blurb in the readme to something in an automated test suite later.

peterbe:
Also, what other type annotations can we start adding to the code do you think?

stereobooster:
other can look like this

```js
  /**
   * @param {string} url
   * @returns {void}
   */
  const addToQueue = url => {}
```

see https://github.com/stereobooster/react-snap/blob/master/src/puppeteer_utils.js for more examples

stereobooster:
as of dev dependency, we can add it as `"typescript": "next"`

peterbe:
So we're not changing to TypeScript? Just using it to do some pre-flight checks? Isn't that what flow does?

stereobooster:
Yes exactly. The difference is that TypeScript has bigger number of type signatures.

peterbe:
What about Flow? Should we consider that as an alternative. 

There's something that doesn't feel right about using TypeScript just to do a "linting" check. Either we should rewrite the code in TypeScript or we should consider Flow or ESLint instead. 

Perhaps I speak too early. Perhaps I just haven't been able to see the power of this in automation.

stereobooster:
> What about Flow? Should we consider that as an alternative.

Only if you willing to write type signatures for puppeteer e.g. this

```diff
+    "@types/node": "^8.0.46",
+    "@types/puppeteer": "^0.12.0",
```

> There's something that doesn't feel right about using TypeScript just to do a "linting" check.

this is how puppeteer project use it. Check their source code

peterbe:
> Only if you willing to write type signatures for puppeteer e.g. this

Sorry for being slow to understand this but what difference does it make what or how puppeteer works? Does using typescript to do *our* linting (and that being the same as puppeteer) mean that it will be able to be aware of the parameters we send into it? Like, both using typescript means it can check that our types when calling puppeteer? And it wouldn't be able to do that otherwise?

> this is how puppeteer project use it. Check their source code

So the original issue is borne out of that? Like, "if they use typescript, we should too"?
But it seems you're saying that not just "inspiration of best practices" but also the powerful benefit that type checks can be made between minimalcss and it calling puppeteer as a lib? If so, cool!

peterbe:
You showed me how to use `tsc` to just run a pass through the source code as is, like I pasted [in this comment](https://github.com/peterbe/minimalcss/pull/22#issuecomment-339377159).
Is it also a possibility to run somehow run `minimalcss` *with* typescript enabled in runtime. Is that a thing or am I completely missing how this all works?

stereobooster:
> Is it also a possibility to run somehow run minimalcss with typescript enabled in runtime. Is that a thing or am I completely missing how this all works?

Sorry, but this is not a thing.

We want TypeScript for static (not dynamic) type checking, e.g. before script executed.

stereobooster:
try to do something like this:

```diff
    page.on('request', request => {
+      console.log(request.something);
```

then

```sh
tsc
src/run.js(38,27): error TS2339: Property 'something' does not exist on type 'Request'.
```

this is why we want typescript

peterbe:
That last comment was impressive. 

Ok. So let's do this! What's needed here is documenting about using `tsc` to do development. 

I vote we come back to this once we have continous integration started. Then we can make a CI rule that checks that no warnings should appear.

stereobooster:
Added documentation

peterbe:
Hurray! 
Now we can start including this in the CI.

stereobooster:
well there are two fails reported, as soon as we remove those - yes we can use CI. Plus we can add unit tests for small things (not puppeteer related). WDYT about Jest?

PR #25, opened 2017-10-25
Ignore images more solidly fixes 24

peterbe:


peterbe:
@stereobooster check it out. 
I made a little example document too. When I run that with `./bin/minimalcss.js http://localhost:8000/images.html` I can see that none of the images get downloaded and the final outcome is:
```css
img{border:1px solid #ccc}p{color:maroon;text-style:italic}
```
which is correct and expected.

Issue #12, opened 2017-09-15
Paths to urls within CSS likely to be busted


peterbe:
If we intercept and download `https://cdn.example.com/prefix/foo.css` and that contains css like this:
```css
.selector {
  background-image: url(../path/static.png);
}
```
If that css (minified and everything) is inserted inline the HTML page, it's going to assume that the file `static.png` exists on the HTML page's domain too. 

Either print a warning. Or attempt to solve it by extrapolating the domain and prefix etc.

stereobooster:
fixed

PR #28, opened 2017-10-25
Resolve CSS URLs fix for #12

stereobooster:


src/run.js
```diff
@@ -242,6 +243,27 @@ const minimalcss = async options => {
...
+              if (value.type === 'Raw') {
+                path = value.value;
+              } else {
+                path = value.value.substr(1, value.value.length - 2);
+              }
+              const absolute = /^https?:\/\/|^\/\//i
+              const root = /^\//
+              if (absolute.test(path)) {
+                // do nothing
+              } else if (root.test(path) /* && css path is the same domain as html document */) {
```
stereobooster: need to fix this

peterbe: Did you want to complete that before I do a full code review?

stereobooster: not sure how to handle this. WDYT?

peterbe:
We really need a unit test framework!

Anyway, consider making an example that uses `<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css">`
That file contains things like `url('../fonts/fontawesome-webfont.eot?#iefix&v=4.7.0')` so once this file is converted - by `minimalcss` - the output should become `url('https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/fonts/fontawesome-webfont.eot?#iefix&v=4.7.0')`.

stereobooster:
Jest? or you prefer something else?

peterbe:
> Jest?

Yes, if you want to get a crack start I'd love to take a look.

stereobooster:
blocking issue https://github.com/Microsoft/TypeScript/issues/7546

stereobooster:
I made it work except this bug in csstree:

```
bin/minimalcss.js http://127.0.0.1:8080/css-relative-path.html
TypeError: node.children.each is not a function
    at Object.eval [as StyleSheet] (eval at createWalker (/minimalcss/node_modules/css-tree/lib/walker/create.js:235:29), <anonymous>:5:15)
    at walk (/minimalcss/node_modules/css-tree/lib/walker/create.js:258:39)
    at Object.walk (/minimalcss/node_modules/css-tree/lib/walker/create.js:264:13)
    at response.text.then.text (/minimalcss/src/run.js:86:21)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

There is an array in one place, while code always expects Lists
```diff
    for (var name in types) {
        if (hasOwnProperty.call(types, name)) {
            var config = types[name];
            walkers[name] = Function('node', 'context', 'walk',
                (config.context ? 'var old = context.' + config.context + ';\ncontext.' + config.context + ' = node;\n' : '') +
                config.fields.map(function(field) {
                    var line = field.type === 'list'
-                        ? 'node.' + field.name + '.each(walk);'
+                        ? 'node.' + field.name + '.forEach ? node.' + field.name + '.forEach(walk) : node.' + field.name + '.each(walk);;'
                        : 'walk(node.' + field.name + ');';

                    if (field.nullable) {
                        line = 'if (node.' + field.name + ') {\n    ' + line + '}';
                    }

                    return line;
                }).join('\n') +
                (config.context ? '\ncontext.' + config.context + ' = old;' : '')
            );
        }
    }
```

stereobooster:
fixed

peterbe:
Great work! But one thing we can maybe fix is to drop the domain name. 
What I get when I run your example is:
```css
background-image:url(http://localhost:8000/images/small.jpg)
```
What it really ought to be is:
```css
background-image:url(/images/small.jpg)
```
because there's no need to repeat the host (and protocol).

stereobooster:
yes you are right. missed it somehow

src/run.js
```diff
@@ -69,19 +70,47 @@ const minimalcss = async options => {
...
+            if (node.type === 'Url') {
+              let value = node.value;
+              let path
+              if (value.type === 'Raw') {
+                path = value.value;
+              } else {
+                path = value.value.substr(1, value.value.length - 2);
+              }
+              const absolute = /^https?:\/\/|^\/\//i
+              const root = /^\//
```
peterbe: These two variables don't need to be variables. They're only used once (each) and they're also a bit "poorly named". Like `absolute` doesn't carry any descriptive value. Unlike "absoluteUrl", "nodeTree" or "tree". Perhaps they should be called "absoluteUrlRegex"?

Can you consider putting the regexes inline?
For example:
```javascript
if (/^https?:\/\/|^\/\//i.test(path)) {
  ...
```

src/run.js
```diff
@@ -69,19 +70,47 @@ const minimalcss = async options => {
...
+              const responseHost = url.parse(responseUrl).host
+              const pageHost = url.parse(pageUrl).host
```
peterbe: This could probably be simplified to just:
```javascript
const sameHost = url.parse(responseUrl).host === url.parse(pageUrl).host
```

stereobooster:
ok, a bit later will do

stereobooster:
I tested bootstrap example and it works. What is not working for you?

peterbe:
> I tested bootstrap example and it works. What is not working for you?

Earlier when I tested with fontawesome.min.css from a CDN, the fonts wouldn't load. But I suspect it's because fonts are weird and I kinda rushed my testing. Let's make this PR be about images for the moment and we can tackle the fonts problem in a separate PR.

PR #29, opened 2017-10-26
option to not ignore images, fixes #26

### LINKED ISSUE #26: Option to actually load images
The default implementation [blocks downloading images](https://github.com/peterbe/minimalcss/blob/0aecdbb7960724e569894b5faca6e8544bbb7d3f/src/run.js#L39) based on the assumption that images being loaded or not don't affect the DOM. Also, the assumption is that downloading images slows things down unnecessarily. 

However, [in this discussion](https://github.com/peterbe/minimalcss/pull/20#issuecomment-339444431) we identified a use case where the image loading does actually affect the DOM. (executing DOM changing code in dynamically inserted images based on `onload`).

The safest thing is to allow for loading images if you believe it's necessary and you know your site actually depends on it.

peterbe:
@stereobooster what do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# dev
+
+* Option to actually load images if you need it to.
+  [#26](https://github.com/peterbe/minimalcss/issues/26)
+
 # 0.2.3
 
 [compare](https://github.com/peterbe/minimalcss/compare/v0.2.2...v0.2.3)
--- a/bin/minimalcss.js
+++ b/bin/minimalcss.js
@@ -51,7 +51,7 @@ if (argv['help']) {
       '  --output <path> or -o <path>  Path to write the final CSS to.\n' +
       '  --verbose                     Include a comment about the options and the date it was generated.\n' +
       '  --debug or -d                 Print all console logging during page rendering to stdout.\n' +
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# dev
```
stereobooster: what is dev stands for? Changelog?

bin/minimalcss.js
```diff
@@ -50,6 +51,7 @@ if (argv['help']) {
       '  --output <path> or -o <path>  Path to write the final CSS to.\n' +
       '  --verbose                     Include a comment about the options and the date it was generated.\n' +
       '  --debug or -d                 Print all console logging during page rendering to stdout.\n' +
+      '  --loadimages                  By default, all images are NOT downloaded. This reversed that.\n' +
```
stereobooster: This reverse~~d~~ that. (without d) I suppose

CHANGELOG.md
peterbe: A marker to remember to change when I update the changelog next time.

bin/minimalcss.js
peterbe: Fixed now.

PR #30, opened 2017-10-26
Generate TS type definitions from jsdoc

stereobooster:
WIP blocked by https://github.com/englercj/tsd-jsdoc/issues/35

### DIFF
--- a/out/types.d.ts
+++ b/out/types.d.ts
@@ -1,0 +1,24 @@
+/**
+ * @param {{urls:Array<string>, debug: boolean, loadimages: boolean}} options
+ * @return {Promise<{ finalCss: string, stylesheetAstObjects:any, stylesheetContents: string }>}
+ */
+declare function minimalcss(options: any): any;
+
+/**
+ * Take call "important comments" and extract them all to the
+ * beginning of the CSS string.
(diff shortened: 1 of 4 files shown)


out/types.d.ts
```diff
@@ -0,0 +1,24 @@
...
+ *   p { background-color: red; }
+ * You can then instead get:
+ *   /*! important 1 *\/
+ *   /*! important 2 *\/
+ *   p { color: red; background-color: red; }
+ * @param {string} css
+ * @return {string}
+ */
+declare function collectImportantComments(css: string): string;
+
```
peterbe: What *is* this file?

stereobooster: this is type definition - if anybody will use this project as library, one will get typescript definitions e.g. you can use it in TypeScript or get intellisense in Code etc.

stereobooster:
> I'm guessing you can figure out the conflict. Was it to do with the new option to the minimize function?

Yes, but anyway this is not ready for merge. It doesn't work as expected.

peterbe:
Exciting! I don't even know how to test that once it's done but it sounds exciting for sure.

stereobooster:
Would be fixed as soon as this will be fixed https://github.com/Microsoft/TypeScript/issues/7546

PR #31, opened 2017-10-26
ability to skip domain in cli, fixes #19

### LINKED ISSUE #19: Prevent network request to Google analytics

peterbe:


peterbe:
@stereobooster What do you think? It's crude but I think it works. 
I went on a big trip of trying to use various wildcard matching libraries. But if you for example want to exclude `a.cdn.foo` *and* `c.cdn.foo` instead of doing `--skip "*.cdn.foo"` you can simply do `--skip .cdn.foo`. 

If someone needs to do much more advanced regular expressions on skips then the cli isn't the right tool.

bin/minimalcss.js
```diff
@@ -73,6 +74,20 @@ const options = {
...
+    let skips = argv['skip']
+    if (!skips) {
+      return false
+    }
+    if (!Array.isArray(skips)) {
+      skips = [skips]
+    }
+    return skips.some(skip => {
+      return request.url.match(
+        new RegExp(skip.replace(/([.*+?^=!:${}()|\[\]\/\\])/g, '\\$1'), 'i')
```
stereobooster: is this for domain extraction? some comment or tests would be helpful

peterbe: I'm glad I asked for a review. No that's just silly to do now. To do case sensitive string comparison we can simply do `url.toLowerCase().indexOf(skip.toLowerCase()) > -1`
Thanks for pointing that out!

peterbe:
@stereobooster 
Check out these new changes. I was testing with a page that had `bulma-docs.css` and `font-awesome.min.css` but I wanted to see what the minimal CSS became if I decided to *skip* `font-awesome.min.css`. E.g. this:

```
▶ ./bin/minimalcss.js --verbose --skip awesome http://peterbecom.dev/css-blocking/bulma-documentation.html
```
If I didn't add this new change, it would eventually fail when it wanted to analyze the `<link rel=stylesheet href=font-awesome.min.css>`

Now it works.

stereobooster:
Makes sense

peterbe:
Good to merge?

Issue #35, opened 2017-10-28
A way to pass browser instance


stereobooster:
In `react-snap` I already have an instance of browser. It would be nice if I can pass it instead of another being initiated by minimalcss. Would you accept PR?

peterbe:
Sure! So, instead of 
```javascript
const browser = await puppeteer.launch({})
```
We'd optionally get it from the `options` object, right?

peterbe:
Strange that the PR didn't automatically close this issue.

PR #36, opened 2017-10-30
Fix for #35

stereobooster:


peterbe:
We need to be better at up-keeping the documentation. Can you please mention this in the README too?

package.json
```diff
@@ -38,7 +38,8 @@
     "typescript": "^2.7.0-dev.20171025"
   },
   "scripts": {
-    "precommit": "lint-staged"
+    "precommit": "lint-staged",
+    "tsc": "tsc -p ."
```
peterbe: Did this slip in from another branch?

stereobooster: nope. decided to add it. So we can `yarn tsc`

stereobooster:
Updated

peterbe:
Again, Thank you so much! I've updated the CHANGELOG too.

PR #37, opened 2017-11-02
userAgent

stereobooster:


src/run.js
```diff
@@ -10,13 +10,14 @@ const url = require('url')
...
  *
- * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any }} options
+ * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string }} options
  * @return Promise<{ finalCss: string, stylesheetAstObjects: any, stylesheetContents: string }>
  */
 const minimalcss = async options => {
   const { urls } = options
   const debug = options.debug || false
   const loadimages = options.loadimages || false
+  const userAgent = options.userAgent || 'minimalcss'
```
peterbe: This makes me nervous. What if the page you're opening is sensitive (ie. full of hacks) that depend on the user agent. 
If you *don't* set it, will it become something Chrome like?
If so, I think we should default to let puppeteer control this unless it's explicitly set. E.g.
```javascript
if (options.userAgent) {
    await page.setUserAgent(options.userAgent)
}
```

If you agree, can you update the options documentation above too?

stereobooster: will do

stereobooster:
@peterbe fixed

peterbe:
0.3.0 now released.

PR #38, opened 2017-11-02
linting script

peterbe:
@stereobooster What do you think?

If there are any files that haven't been run through `prettier` it looks like this:
```
▶ yarn lintcheck
yarn run v1.2.1
warning ../package.json: No license field
warning ../../../package.json: No license field
$ prettier --list-different "src/*.js" "bin/*.js"
src/run.js
src/utils.js
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Perhaps it'd be more useful to move this to a bash script instead of trying to squeeze this all into `package.json`. Then, if it does list any it could print an additional message about what you need to do to fix the problems.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,2 @@
+semi: false
+singleQuote: true
--- a/README.md
+++ b/README.md
@@ -91,8 +91,7 @@ It can optionally use `PhantomJS` to extract the HTML.
 This is highly experimental.
 
 The goal is to expand the tooling with all the bells and whistles that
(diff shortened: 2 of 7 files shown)


stereobooster:
you want to use it in CI?

peterbe:
> you want to use it in CI?

Indeed. Right now it'll just be forgotten but once part of CI it'll be good.

peterbe:
This is a start. At least it's there. We can wire it up to CI later.

PR #42, opened 2017-11-15
Upgrade puppeteer

stereobooster:


src/run.js
```diff
@@ -90,7 +90,7 @@ const minimalcss = async options => {
       throw error
     })
 
-    const response = await page.goto(url, { waitUntil: 'networkidle' })
+    const response = await page.goto(url, { waitUntil: 'networkidle2' })
```
stereobooster: not sure about this. Do we want networkidle0 or networkidle2

src/run.js
```diff
@@ -120,7 +120,7 @@ const minimalcss = async options => {
       throw error
     })
 
-    const response = await page.goto(pageUrl, { waitUntil: 'networkidle' })
+    const response = await page.goto(pageUrl, { waitUntil: 'networkidle2' })
```
stereobooster: not sure about this. Do we want networkidle0 or networkidle2

PR #53, opened 2017-11-28
evaluate dom multiple times, fixes #52

### LINKED ISSUE #52: Evaluate DOM multiple times
Currently, we're [doing this](https://github.com/peterbe/minimalcss/blob/4dead07e4aa8eef429124577be1b0ba9799a8549/src/run.js#L123) which means we load the page, wait for [`networkidle2`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagegotourl-options).

What *could* happen is that the app needs certain CSS **before** the whole page is loaded. E.g. a `<div id="loading-animation">` DOM node only appears in the DOM for the first milliseconds before the JavaScript kicks in and changes the things. Let's call that placeholder DOM. Animations and "Please wait..." elements are such examples. These's CSS should be part of the minimal CSS.

peterbe:
Hey @stereobooster What do you think about this?
The diff is really large so it might be hard to see what's going on. But basically...

It does this:
```javascript
const browser = await puppeteer.launch({})
for (url in urls) {
  const page = await browser.newPage()
  await page.goto(url)
  const html1 = await page.evaluate(() => {
    return document.documentElement.outerHTML;
  })
  await page.goto(url, {waitUntil: 'networkidle2'})
  const html2 = await page.evaluate(() => {
    return document.documentElement.outerHTML;
  }) 
  ...Note that we have two strings of HTML called 'html1' and 'html2'
}
browser.close()
```
So, in the `page.evaluate` the only thing we do is that we send the `document.body` DOM element as a string in HTML. 
Once that's done, use [cheerio](https://www.npmjs.com/package/cheerio) to parse the HTML strings (if you're not familiar with cheerio, it's basically like jQuery but for node)

Once all CSS and all HTML has been downloaded you're outside Puppeteer and can post process it all in full Node without having to do this work inside the `page.evaluate` callback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+# 0.4.0
+
+* Every URL you pass gets loaded twice. First *without Javascript* and then
+  *with JavaScript* (and waiting for network to be idle). These means the
+  minimal CSS will contain CSS that was necessary **before** the page is fully
+  loaded as well.
+  Also, the engine has entirely changed. Instead of evaluating the DOM inside
+  a page evaluation (the equivalent of running in the Web Console), puppeteer
+  is only used to 1) download relevant assets and 2) yield the DOM as a string
+  of HTML. Instead [cheerio](https://www.npmjs.com/package/cheerio) is used
+  to compare the CSS to the DOM.
+  [pull#53](https://github.com/peterbe/minimalcss/pull/53)
+
 # 0.3.1
 
 * Any errors raised internally by `document.querySelector()` are not
--- a/README.md
+++ b/README.md
@@ -155,6 +155,20 @@ key is `urls`. Other optional options are:
 * `browser` - Instance of a [Browser](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-browser), which will be used instead of launching another one.
 * `userAgent` - specific user agent to use (string)
 
+## Warnings
+
+When `minimalcss` evaluate each CSS selector to decide whether to keep it
+or not, some selectors might not be parseable. Possibly, the CSS employs
+hacks for specific browsers that
+[cheerio](https://www.npmjs.com/package/cheerio) doesn't support. Or
+there might be CSS selectors that no browser or tool can understand
+(e.g a typo by the CSS author). If there's a problem parsing a CSS selector,
+the default is to swallow the exception and let the CSS selector stay.
+
+Also by default, all these warnings are hidden. To see them use the `--debug`
+flag (or `debug` API option). Then the CSS selector syntax errors are
+printed on `stderr`.
+
 ## Development
 
 We use ES6+ with jsdoc comments and TypeScript to do static type checking, like [puppeeteer does](https://github.com/GoogleChrome/puppeteer/pull/986/files).
--- a/README.md
+++ b/README.md
@@ -147,7 +147,11 @@ key is `urls`. Other optional options are:
   as an argument and returns boolean. If it returns true then given request
   will be aborted (skipped). Can be used to block requests to Google Analytics
   etc.
-* `loadimages` - If set to true, images will actually load.
+* `loadimages` - If set to `true`, images will actually load.
+* `withoutjavascript` - If set to `true` it will *skip* loading the page first
+  without JavaScript. By default `minimalcss` will evaluate the DOM as plain as
+  can be, and then with JavaScript enabled *and* waiting for network activity
+  to be idle.
 * `browser` - Instance of a [Browser](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-browser), which will be used instead of launching another one.
 * `userAgent` - specific user agent to use (string)
 
--- a/bin/minimalcss.js
+++ b/bin/minimalcss.js
@@ -18,7 +18,8 @@ const argv = minimist(args, {
     'version',
     'verbose',
     'debug',
-    'loadimages'
+    'loadimages',
+    'withoutjavascript'
   ],
   string: ['output', 'skip'],
   default: {
@@ -52,6 +53,8 @@ if (argv['help']) {
       '  --verbose                     Include a comment about the options and the date it was generated.\n' +
       '  --debug or -d                 Print all console logging during page rendering to stdout.\n' +
       '  --loadimages                  By default, all images are NOT downloaded. This reverses that.\n' +
+      '  --withoutjavascript           The CSS is evaluated against the DOM twice, first with no JavaScript, ' +
+      'then with. This disables the load without JavaScript.\n' +
       '  --skip                        String to match in URL to ignore download. Repeatable. E.g. --skip google-analyics.com\n' +
       '  --version or -v               Print minimalcss version.\n' +
       ''
@@ -74,6 +77,7 @@ const options = {
   urls: urls,
   debug: argv['debug'],
   loadimages: argv['loadimages'],
+  withoutjavascript: argv['withoutjavascript'],
   skippable: request => {
     let skips = argv['skip']
     if (!skips) {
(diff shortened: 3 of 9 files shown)


peterbe:
**So, how is this different?**

Please check out [the issue](https://github.com/peterbe/minimalcss/issues/52).

Basically, in the `examples/stages.html` example, the DOM mutates after the XHR request has finished. It basically removes the `initial` class name from the `<p>` tag. 

**Things to consider**

It currently looks at the DOM after the HTML and the `<script>` tags have been evaluated. Meaning, it includes the DOM *after* `window.onload` has fired. As of this pull request, that is solved.

Perhaps we should consider the DOM before *any* JavaScript has been evaluated at all. 

Also, perhaps it's crazy to consider the DOM **after** the XHR requests have finished. 


What do you think?

**Lastly**
I have not tested this new code yet. Just that `examples/stages.html` page so far.

stereobooster:
1. You can get html with `let content = await page.content();`.
2. You can get initial html with nodejs `http.get({...` 

I'm not sure about this case - this is kind of anti-pattern, to mutate DOM on load (without user interaction). This will always create waterfall chain of requests, so longer time to load. What the sense of using "critical" CSS if you have another performance bottleneck...

stereobooster:
When I say "to mutate DOM" - I mean to provide any DOM for the browser which will immediately change after onload. The browser just finished to paint everything, and you say do some changes - so more reflows, more paints.
In case of React, typical setup is empty HTML documents. And you do not mutate DOM, you basically create it from scratch.

peterbe:
Note that there are 3 repaints in `examples/stages.html`. To see it you need to really slow down your browser with dev tools and watch carefully. 
And I don't think it's an unrealistic thing, to have some CSSOM to display to the user whilst waiting for the JavaScript to execute and even some more CSSOM whilst waiting for XHR to finish loading.

peterbe:
I'm going to investigate the alternative of using `let content = await page.content();.`.

stereobooster:
> And I don't think it's an unrealistic thing, to have some CSSOM to display to the user whilst waiting for the JavaScript to execute and even some more CSSOM whilst waiting for XHR to finish loading.

Yes this is totally possible. I do understand what you are talking about. I'm saying that those websites have bigger problems in sense on loading performance optimisation. So they need to fix this first and then switch to critical CSS.

If you are going to invest time to integrate pretty complex solution with critical css, but haven't fixed easier problems in first place `¯\_(ツ)_/¯`.

peterbe:
Another use case is that you might NOT depend on XHR to load the main content. But perhaps, the site has a side-bar or something that starts loading content immediately after the initial page load. You know, just to break up the load. Or because that stuff in the sidebar isn't important to SEO. 

Here's [a site of mine](https://dinnerd.peterbe.com) that is a single-page-app. As soon as the JavaScript has been downloaded and React starts, it starts to draw a DOM. Once that's done (`componentDidMount`) it connects to Firebase to download the actual user data. 
It [becomes apparent if you slow down the loading](https://www.webpagetest.org/video/compare.php?tests=171201_A4_d5bcf530d6c1d3a85a64d8b8f8a80d0d-r:1-c:0).

Whatever, the use case is real and we should solve for it. My prototype is promising I believe. The only thing is, should we analyze the DOM in that middle case. I.e. after the JavaScript has loaded but before the XHR have finished. ?????

This PR currently does all three. Before Javascript, After Javascript and After XHR. 
In all fairness, that's the safest thing to do. But requires to trigger 3 page loads in puppeteer instead of just 2.

stereobooster:
From my POV you are overcomplicating. Isn't CSS will be loaded by the moment when React will start to execute? (we are talking about critical CSS + load rest of it async right?)

What is the size of CSS? 20 kb gzip
What is the size of JS with react? 60 kb gzip + time to parse and compile

I'm not against your idea. If you want go for it. I'm just saying that you are increasing complexity for doubtful case

Disclaimer: I haven't checked your example, a bit busy right now. Judging based on my previous experience. Actually need to measure with webpagetest to argue on this

stereobooster:
On the second thought. Two cases make sense:
- CSS without JS
- CSS with JS at the end

Intermediate point is a bit over complication from my POV, plus there is guarantee that you will catch "exact moment".

---

As of given example  I would go for other improvements before critical CSS. I would not use appcache directly, but only with appcache-nany. Plus I would make system to chose service-workers if it is available and fallback to appcache-nany. Use async in scripts.

screen shot 2017-12-01 at 23 12 21

peterbe:
This is tangential but the argument for using appcache is strong. It sucks compared to service workers and you get a terrible score on Chrome Lighthouse, but if you want to make a good offline'ish web-page-as-an-app on iOS, the appcache is the only tool available. Without it, opening a web page on iOS (once made into an icon on the Home screen) will always have to wait for a white screen whilst it basically reloads the app.

peterbe:
Long story short, I'm going to amend this PR to...

1) Only use: Before JS, and After JS and network idle.
2) Use `page.content` instead to get the HTML instead of the current hack.

stereobooster:
About offline. Here is how I'm doing https://github.com/stereobooster/an-almost-static-stack/blob/react-snap/src/registerServiceWorker.js

```js
//check for serviceworker first
if ("serviceWorker" in navigator) {

// if not available check for applicationCache
} else if (window.applicationCache) {

}
```

this way you will get serviceWorker and applicationCache as fallback

peterbe:
@stereobooster It bothers me how long this has been sitting stale. It's time to increase the pace!

I went back, did some cleaning and now I have something that I think works really well. 
It does what [you mentioned](https://github.com/peterbe/minimalcss/pull/53#issuecomment-348628668).

* CSS without JS
* CSS with JS at the end

In the examples folder I have a `stages.html` sample which has some CSS and DOM that gets rendered first but gets "overwritten" by JavaScript later when it loads. That passes now. 

I'm still hesitant that we might be missing an important use case. Suppose you have a piece of code like this:
```javascript
$(function() {  // on page load
  $('#content').append($('<p>').addClass('loading-paragraph'));
  $('#content').get('/server/getting/html');
})
```
In this case you have a CSSOM that only exists AFTER JavaScript has loaded and BEFORE the XHR has finished. `minimalcss` would fail this. Your point is valid that this is unusual and perhaps there are better ways to solve this. (For example https://dinnerd.peterbe.com puts in a loading message in the `<div id="app">` DOM whilst waiting for React to load and that DOM is [styled with inline tags](https://github.com/peterbe/dinnerd/blob/e001ffb75cce7f112ec53a589753bfb0097568f7/public/index.html#L45-L55))

To avoid that whole discussion again, I decided to ditch it for now and left it in as [a code comment](https://github.com/peterbe/minimalcss/pull/53/files#diff-f712f2169bd3dca93334576e35acff9fR145).

peterbe:
This PR is quite big. It totally changes how `minimalcss` works. Instead of doing all the magic of comparing CSS selectors to the DOM inside the page evaluation, it now only uses puppeteer to extract the DOM as a HTML string and all the other core functionality is now done in Node with `cheerio`. 

Please keep an eye out to see if it breaks things.

peterbe:
In release [0.4.0](https://github.com/peterbe/minimalcss/releases/tag/v0.4.0).

stereobooster:
>  the DOM as a HTML string and all the other core functionality is now done in Node with cheerio

I see potential improvement here. We can extract the core functionality and have it without puppeteer functionality.

stereobooster:
Oh and this "puppeteerless" core is pretty easy to test with unit tests

peterbe:
> I see potential improvement here. We can extract the core functionality and have it without puppeteer dependency.

Yes! I'm assuming you mean to rearrange the code so that we have two distinct modules. One for networky stuff and one for dealing with arrays of HTML blobs. 

I got started with that `src/utils.js` to try to break out some generic stuff to make the `src/run.js` smaller but then I think I got distracted. Please have a go if you like. We can do that first and then worry about jest and travis and stuff.

PR #56, opened 2017-12-25
cheerio types

stereobooster:
I want to refactor code to have puppeteerless core, but puppeteer is deeply entangled in current code. This is all I managed to do so far :rofl:

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "puppeteer": "^0.13.0"
   },
   "devDependencies": {
+    "@types/cheerio": "^0.22.6",
     "@types/node": "^8.0.57",
     "@types/puppeteer": "^0.13.7",
     "husky": "^0.14.3",
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "csso": "3.4.0",
     "filesize": "^3.5.11",
     "minimist": "^1.2.0",
+    "node-fetch": "^1.6.7",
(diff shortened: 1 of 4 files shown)


peterbe:
Great start! I like the name `extract.js`, however, perhaps we can do even better. Keep `run.js` and make it really small. Then it would called `download.js` and later you pass its results to `extract.js`. 

I haven't dug too deep into the diff yet but one immediate thing that strikes me is why would the refactoring require `fetch()`? puppeteer should do all the necessary network traffic needed.

stereobooster:
> puppeteer should do all the necessary network traffic needed.

I use puppeteer to render a webpage, everything else is handled purely by node.js. Not saying this is the final version, rather was thinking out loud.

peterbe:
I think it's much safer and saner to use puppeteer to do both rendering AND using it to figure out which CSS it actually needs to download.

Because this isn't ready yet, do you mind if I land https://github.com/peterbe/minimalcss/pull/57? I'm excited to use that to also clean up un-used font-families.

stereobooster:
> Because this isn't ready yet, do you mind if I land #57?

Go for it 👍

peterbe:
This now conflicts. The diff might be really confusing. If you want to I would welcome a break-up of this PR. One just for the cheerio tsc types. And one that just breaks up `run.js` a little so it's doing fewer things. I'm sure there are functions and code in it that far from the puppeteer instance.

PR #57, opened 2017-12-28
strip unused keyframes, fixes #14

### LINKED ISSUE #14: Strip unused keyframe animations
If you have a keyframe, like:
```css
@keyframes cssload-global {
  0%, 100% {
    transform: rotateY(0deg);
  }
  50% {
    transform: rotateY(360deg);
  }
}
```

We should only keep it if there is a thing like this:
```css
animation: 2.3s cssload-global infinite;
```
...for example. 

Note that we can't just look for `animation`. We might also need to look for things like `-ms-animation`.

peterbe:
@stereobooster This works. There might be a better way to do it but it actually works. 
Contrast the [CSS used in the example](https://github.com/peterbe/minimalcss/blob/dce1c8eb07e7eba3043c2defa21bc1799ccf4d69/examples/animations/index.css) with the output you get when you run this:
```css
@keyframes App-logo-spin {
    0% {
        transform: rotate(0deg)
    }
    to {
        transform: rotate(360deg)
    }
}

@keyframes pulse {
    0% {
        background-color: #fff5f5;
        color: #000
    }
    to {
        background-color: #8ac3ff;
        color: #eee
    }
}

.loaded, .loading {
    margin-top: 100px;
    text-align: center
}

@media only screen and (min-device-width:768px) and (max-device-width:1024px) and (-webkit-min-device-pixel-ratio:1) {
    .loaded, .loading {
        margin-top: 50px
    }
}

.loading {
    animation: App-logo-spin infinite 5s linear;
    height: 80px
}

.loaded {
    animation-name: pulse;
    animation-duration: 3s;
    animation-iteration-count: infinite;
    animation-timing-function: ease-out
}
```

Basically, it was able to figure out that there is never a need for the animations `stretch` and `progress-bar-stripes` so they get cleaned out.

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -64,7 +64,7 @@ const postProcessKeyframes = ast => {
 
 /**
  *
- * * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean }} options
+ * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean }} options
  * @return Promise<{ finalCss: string, stylesheetAstObjects: any, stylesheetContents: string }>
  */
 const minimalcss = async options => {
--- a/src/run.js
+++ b/src/run.js
@@ -168,17 +168,6 @@ const minimalcss = async options => {
                 }
               }
               value.value = path
-              // } else if (
-              //   node.type === 'Declaration' &&
-              //   node.property.search(/animation$/) > -1
-              // ) {
-              //   let firstName = false
-              //   node.value.children.each(child => {
-              //     if (child.type === 'Identifier' && child.name && !firstName) {
-              //       allAnimationNames.add(child.name)
-              //       firstName = true
-              //     }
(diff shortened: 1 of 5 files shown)


peterbe:
Another promising example of this is:
```
▶ ./bin/minimalcss.js -d -o /tmp/output.master.css https://www.peterbe.com/search\?q\=found
```
vs. (switching into this branch)
```
▶ ./bin/minimalcss.js -d -o /tmp/output.keyframesremoved.css https://www.peterbe.com/search\?q\=found
```
And the result is promising:
```
/tmp
▶ ls -ltrh *.css
-rw-r--r--  1 peterbe  wheel    56K Dec 27 20:43 output.keyframesremoved.css
-rw-r--r--  1 peterbe  wheel    81K Dec 27 20:43 output.master.css
```
That's **25KB less CSS**. (granted that difference shrinks when you gzip)

src/run.js
```diff
@@ -116,6 +168,17 @@ const minimalcss = async options => {
                 }
               }
               value.value = path
+              // } else if (
```
stereobooster: remove commented code

PR #58, opened 2017-12-29
Testcase for css-in-js

stereobooster:
Related to https://github.com/stereobooster/react-snap/issues/76

All possible examples of inserting CSS into the document. As of now it only tracks external files

```
./bin/minimalcss.js http://127.0.0.1:8080/css-in-js.html
.external{color:red}
```

The question is: should `minimalcss` track css-in-js generated css or not. ~~`minimalcss` should not take into account existing style tags (`.inline` in given example).  They can be tracked at first load (without JS).~~

UPD: if we treat `minimalcss` as a tool which returns set of css rules required to render current page it doesn't matter where css comes from. It would be nice to have config to specify what to take into account: external (e.g. `<link>`), dynamic styles, all available styles. WDYT?

### DIFF
--- a/examples/css-in-js.css
+++ b/examples/css-in-js.css
@@ -1,0 +1,1 @@
+.external { color: red }
--- a/examples/css-in-js.html
+++ b/examples/css-in-js.html
@@ -6,6 +6,16 @@
     <style type="text/css">
       .inline { color: red }
     </style>
+  </head>
+  <body>
+    <p class="external">external</p>
+    <p class="inline">inline</p>
+    <p class="cssinjs1">cssinjs1</p>
+    <p class="cssinjs2">cssinjs2</p>
+    <p class="cssinjs3">cssinjs3</p>
+    <p class="cssinjs4">cssinjs4</p>
+    <p class="cssinjs5">cssinjs5</p>
+    <p class="cssinjs6">cssinjs8</p>
     <script type="text/javascript">
       (() => {
         const css = '.cssinjs1 { color: red; }',
@@ -33,11 +43,18 @@
         sheet.insertRule(css, sheet.cssRules.length)
       })();
     </script>
-  </head>
-  <body>
-    <p class="external">external</p>
-    <p class="inline">inline</p>
-    <p class="cssinjs1">cssinjs1</p>
-    <p class="cssinjs2">cssinjs2</p>
+    <script type="text/javascript">
+      const blob = new Blob(['.cssinjs3 { color: red; }'], {type: 'text/css'});
+      const link = document.createElement('link');
+      link.rel = 'stylesheet';
+      link.href = window.URL.createObjectURL(blob);
+      const head = document.head || document.getElementsByTagName('head')[0];
+      head.appendChild(link);
+    </script>
+    <script type="text/javascript">
+      document.querySelector('.cssinjs4').style.cssText = "color: red";
+      document.querySelector('.cssinjs5').setAttribute("style", "color:red");
+      document.querySelector('.cssinjs6').style.color = "red";
+    </script>
   </body>
 </html>
--- a/examples/css-in-js.html
+++ b/examples/css-in-js.html
@@ -1,0 +1,43 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <link href="css-in-js.css" rel=stylesheet>
+    <style type="text/css">
+      .inline { color: red }
+    </style>
+    <script type="text/javascript">
+      (() => {
+        const css = '.cssinjs1 { color: red; }',
+            head = document.head || document.getElementsByTagName('head')[0],
+            style = document.createElement('style');
+
+        style.type = 'text/css';
+        if (style.styleSheet){
+          style.styleSheet.cssText = css;
+        } else {
+          style.appendChild(document.createTextNode(css));
+        }
+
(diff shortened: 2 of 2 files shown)


stereobooster:
I solved original issue in `react-snap`. Not sure if `minimalcss` should handle those cases or not. Let me know what do you think and close PR

peterbe:
When you say "css-in-js" I think about all those react libs that help you generate `style="..."` attributes on tags. Those are impossible to optimize. 

[My previous attempt at all of this](https://github.com/peterbe/mincss) didn't really have a cli. It just returned an object that contained, a minimal CSS block *for each* external CSS **and** *for each* inline style block. The original design with `mincss` was to return the CSS almost as a diff so you could go back to your CSS (inline blocks or .css files) and change them. Now, I think that's generally a bad idea because the true meaning of all this is usually to make that `bootstrap.min.css` (or whatever) smaller. 

So what if we add an option to `minimalcss` so that it spits out:
```css
.external {...}
.inline {...}
.css-in-js {...}
.css-in-js2 {...}
```
(to match your example file).
What are you going to do with that? 

If you change your `index.html` to just inline all the minimal CSS, then don't you need to also change the JS code to no longer do what it did, no?

However, it certainly would be nice to have a way to do something like `./bin/minimalcss.js --include-inline ...` and then replace all `<link rel=stylesheet>` *and* all `<style>` blocks with just 1 `<style>` block. But we can't really rewrite how the JS works. So perhaps the output should only be...:
```css
.external {...}
.inline {...}
```

stereobooster:
> When you say "css-in-js" I think about all those react libs that help you generate `style="..."` attributes on tags

Actually most of them generate `<style></style>` tags instead of inline style attribute. Inline style is in minority, but actually thank you for reminding existence of that one.

> What are you going to do with that?

Have no idea.

Thank you for response. I will need to boil this down in my head.

peterbe:
I actually thing we should keep this open. The use case isn't a top priority but it's still something we have a great opportunity to solve. Perhaps:
```
$ ./bin/minimalcss.js --inlines URL
```
And it could *just* download each `style` block, using `cheerio`, and compare it against the DOM(s). We're really close to being able to do this.

Also, since we have access to the `document` in puppeteer we have access to `document.styleSheets`.

The question is, what should the "API" (and cli parameters) be for this?

* We could **combine** the inline blocks with all the downloaded external stylesheets into one blob of CSS.
* We could make it ONLY analyze the inline blocks. 
* (Primarily only a concern for the cli) We could make it pretty-print each inline block with a comment like `/* Inline block number 3 */`

Perhaps we start with a basic implementation and worry about the CLI later. E.g.
```javascript
minimalcss
  .minimize({ urls: ['http://peterbecom.dev/css-blocking/ultra-basic.html'], inlines: true })
  .then(result => {
    result.inlines.forEach((block, i) => {
       console.log(`/* Inline style block number ${i + 1} */`)
       console.log(block.finalCss)
       console.log(`/* Used to be ${block.stylesheetContent.length}bytes but is now ${block.finalCss.length} */`)
    })
  })
```

If a site has (and this is not unusual) 1 external `somefatframework.css` and an inline style block, it's extremely unlikely that the bytes savings on the inline style block will be impressive (or even worth it!) but it's nice to see if you have styles in an inline style block that is now never ever used and can be cleaned up.

stereobooster:
For the fullness of examples: there are also style attribute (we can ignore this) and `Blob`

peterbe:
The `Blob` thing we can do something about. If we can access `document.styleSheets` in the DOM we can access `document.getElementsByTagName('link')`.

The style stuff I don't know you mentioned that.

stereobooster:
branch can be safely merged, because it adds to example folder

peterbe:
Do you want to wait till we have some code that actually does the minimal css work on the inline style blocks?

stereobooster:
We can wait. On the other side this is mergeable and I do not like long standing PRs (I mean in my repositories), but this is up to you

peterbe:
One reason it might be very important to look into every single inline CSS is because of this: https://github.com/peterbe/minimalcss/pull/81

Suppose you have a `foo.css` that contains...
```css
@font-face {
  font-family: 'Rating';
  src: url(data:application/x-font-ttf;charset=utf-8;base64,AAEAAAALAIAAAwAwT1MvMggjCBsAAAC8AAAAYGNtYXCj2p ....
```
BUT in the `foo.css` there is NO:
```css
body { font-family: Rating }
```
So `minimalcss` would think there's no CSS selector with this font-face mentioned. 
However, you might have, in the HTML:
```html
<link rel="stylesheet" href="foo.css">  <!-- defines all font-faces -->
<style>
body { font-family: Rating }
</style>
```

peterbe:
At the moment, we generate [a big AST object](https://github.com/peterbe/minimalcss/pull/81/files#diff-f712f2169bd3dca93334576e35acff9fR394)  called `allCombinedAst`.
If we could have an AST object for ALL external and ALL inline CSS, we could send that into that function and basically tell it to look for *use* of the font-face in each and every AST object.

stereobooster:
test case itself in the master. Let's close PR and open ticket with this https://github.com/peterbe/minimalcss/pull/58#issuecomment-356730109 ?

peterbe:
I filed https://github.com/peterbe/minimalcss/issues/114 to pick up that other interesting discussion.

Issue #50, opened 2017-11-20
option for viewport


stereobooster:
https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagesetviewportviewport

stereobooster:
resolved

PR #64, opened 2018-01-06
[#50] option for viewport

stereobooster:


peterbe:
That's awesome but we need to remember to always document. Can you please add something to the README about this option. 

Also, shouldn't this be possible from the cli too? I appreciate that that *could* be a separate PR later but it's probably a good idea to include that now so we don't forget.

Issue #60, opened 2017-12-30
Error handling


stereobooster:
Related https://github.com/stereobooster/react-snap/issues/45

Error thrown inside promise is not handled. Probably need catch it and expose as `Promise.reject`

peterbe:
Where is it not handled? Also, isn't it a good thing that a broken JS code blows up because if JS isn't working on the page, how can `minimalcss`, with confidence, know what CSS is needed.

stereobooster:
> Where is it not handled? 

`react-snap` as user of `minimalcss` can not take over error handling

peterbe:
Let's fix that!

stereobooster:
resolved

PR #65, opened 2018-01-06
[#60] Proper error handling

stereobooster:
Before
```
./bin/minimalcss.js http://127.0.0.1:8080/jserror.html
(node:7147) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Error: unhandled
    at http://127.0.0.1:8080/jserror.html:7:19
(node:7147) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

After
```
./bin/minimalcss.js http://127.0.0.1:8080/jserror.html
Error: Error: unhandled
    at http://127.0.0.1:8080/jserror.html:7:19
    at Page._handleException (~/minimalcss/node_modules/puppeteer/lib/Page.js:370:38)
    at Session.Page.client.on.exception (~/minimalcss/node_modules/puppeteer/lib/Page.js:95:60)
    at emitOne (events.js:115:13)
    at Session.emit (events.js:210:7)
    at Session._onMessage (~/minimalcss/node_modules/puppeteer/lib/Connection.js:210:12)
    at Connection._onMessage (~/minimalcss/node_modules/puppeteer/lib/Connection.js:105:19)
    at emitOne (events.js:115:13)
    at WebSocket.emit (events.js:210:7)
    at Receiver._receiver.onmessage (~/minimalcss/node_modules/ws/lib/WebSocket.js:143:47)
    at Receiver.dataMessage (~/minimalcss/node_modules/ws/lib/Receiver.js:389:14)
```

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -105,128 +105,148 @@ const minimalcss = async options => {
       }
     })
 
-    // To build up a map of all downloaded CSS
-    page.on('response', response => {
-      const responseUrl = response.url()
-      const ct = response.headers()['content-type'] || ''
-      if (!response.ok()) {
-        throw new Error(`${response.status()} on ${responseUrl}`)
+    await new Promise(async (resolve, reject) => {
+      // If anything goes wrong, for example a `pageerror` event or
+      // a bad download request (e.g. !response.ok), then remember that
+      // we have fulfilled the promise and don't want to call `reject` or `resolve`
+      // a second time.
+      let fulfilledPromise = false
+
+      const safeReject = error => {
+        if (!fulfilledPromise) {
+          reject(error)
+        }
       }
-      if (ct.indexOf('text/css') > -1 || /\.css$/i.test(responseUrl)) {
-        response.text().then(text => {
-          const ast = csstree.parse(text)
-          csstree.walk(ast, node => {
-            if (node.type === 'Url') {
-              let value = node.value
-              let path
-              if (value.type === 'Raw') {
-                path = value.value
-              } else {
-                path = value.value.substr(1, value.value.length - 2)
-              }
-              const sameHost =
-                url.parse(responseUrl).host === url.parse(pageUrl).host
-              if (/^https?:\/\/|^\/\//i.test(path)) {
-                // do nothing
-              } else if (/^\//.test(path) && sameHost) {
-                // do nothing
-              } else {
-                const resolved = new url.URL(path, responseUrl)
-                if (sameHost) {
-                  path = resolved.pathname
+
+      // To build up a map of all downloaded CSS
+      page.on('response', response => {
+        const responseUrl = response.url()
+        const ct = response.headers()['content-type'] || ''
+        if (!response.ok()) {
+          return safeReject(new Error(`${response.status()} on ${responseUrl}`))
+        }
+        if (ct.indexOf('text/css') > -1 || /\.css$/i.test(responseUrl)) {
+          response.text().then(text => {
+            const ast = csstree.parse(text)
+            csstree.walk(ast, node => {
+              if (node.type === 'Url') {
+                let value = node.value
+                let path
+                if (value.type === 'Raw') {
+                  path = value.value
                 } else {
-                  path = resolved.href
+                  path = value.value.substr(1, value.value.length - 2)
                 }
+                const sameHost =
+                  url.parse(responseUrl).host === url.parse(pageUrl).host
+                if (/^https?:\/\/|^\/\//i.test(path)) {
+                  // do nothing
+                } else if (/^\//.test(path) && sameHost) {
+                  // do nothing
+                } else {
+                  const resolved = new url.URL(path, responseUrl)
+                  if (sameHost) {
+                    path = resolved.pathname
+                  } else {
+                    path = resolved.href
+                  }
+                }
+                value.value = path
               }
-              value.value = path
-            }
+            })
+            stylesheetAsts[responseUrl] = ast
+            stylesheetContents[responseUrl] = text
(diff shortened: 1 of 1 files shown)


src/run.js
```diff
@@ -75,283 +75,293 @@ const minimalcss = async options => {
...
+        allHrefs.add(href)
+      })
 
-    // We can close the browser now that all URLs have been opened.
-    if (!options.browser) {
-      browser.close()
+      // We can close the browser now that all URLs have been opened.
+      if (!options.browser) {
+        browser.close()
+        browser = null
```
peterbe: This needs a comment that explains **why** the `browser` variable is changed.

src/run.js
```diff
@@ -75,283 +75,293 @@ const minimalcss = async options => {
...
+    csstreeAst = postProcessKeyframes(csstreeAst)
+    finalCss = csstree.translate(csstreeAst)
 
-  const returned = { finalCss, stylesheetAstObjects, stylesheetContents }
-  return Promise.resolve(returned)
+    const returned = { finalCss, stylesheetAstObjects, stylesheetContents }
+    resolve(returned)
+  }).catch(function(reason) {
+    if (!options.browser && browser) {
+      browser.close()
```
peterbe: Why is there no `browser = null` here?

src/run.js
stereobooster: I suppose there is better way to do this, what is basically required is `finally` to make sure we close browser, instead I'm using `browser=null` as a flag that browser already closed. Probably need to fix this

peterbe:
I'm not sure I'm testing this right. But I'm tempted to land this as a good start. 
The ideal thing would be that the code inside the `catch()` runs "immediately" and then the whole program exist also "immediately".

stereobooster:
I will look into this 30s problem. Do not merge just yet

stereobooster:
```js
page.on('pageerror', error => {
  throw error;
})
```

This doesn't work because an error is thrown outside of the stack.


```js
page.on('pageerror', error => {
  return reject(error)
})
```

This doesn't work because we reject Promise, but do not stop the main function - eventually, main function will finish and Promise will be rejected again or accepted, which is wrong.

We can do something like this:

```js
let pageerror;

page.on('pageerror', error => {
  pageerror = error
})

...

if (pageerror) {
  return Promise.reject(pageerror)
} else {
  return Promise.resolve(...)
}
```

stereobooster:
Or we can convert pagerror to event stream and convert crawling process to stream, merge both streams, reduce streams to promise, as soon as one of the streams fail we will reject Promise.

peterbe:
> Or we can convert pagerror to event stream and convert crawling process to stream, merge both streams, reduce streams to promise, as soon as one of the streams fail we will reject Promise.

Uhhh? I understand about 0.1% of that :)

stereobooster:
This is because I overcomplicated it. Pushed new version. Now seems to work as expected

src/run.js
```diff
@@ -211,8 +219,9 @@ const minimalcss = async options => {
         waitUntil: ['domcontentloaded', 'networkidle2']
       })
       if (!response.ok) {
-        return reject(
-          new Error(`${response.status} on ${pageUrl} (second time)`)
+        return (
+          !pageerrorCalled &&
+          reject(new Error(`${response.status} on ${pageUrl} (second time)`))
```
peterbe: Can you move this, and the block on line 194-197 to a function? Perhaps something like:
```javascript
if (!response.ok) {
  return rejectResponse(response, 'second time')
}
```
and then that function could be something like 
```javascript
const rejectResponse = (response, message) => {
  if (!pageerrorCalled) {
     // We have not previously errored out before, so do it for this bad response
    const error = new Error(`${response.status} on ${pageUrl} (${message})`)
    reject(error)
  }
}
```

src/run.js
```diff
@@ -177,7 +177,12 @@ const minimalcss = async options => {
     })
 
     await new Promise(async (resolve, reject) => {
-      page.on('pageerror', error => reject(error))
+      let pageerrorCalled = false
```
peterbe: The name isn't great. The *point* of this variable is to **avoid calling `reject()` or `resolve()` more than once**, right? How about something like this:

```javascript
// If anything goes wrong, for example a `pageerror` event or 
// a bad download request (e.g. !response.ok), then remember that 
// we have fulfilled the promise and don't want to call `reject` or `resolve`
// a second time.
let fulfilledPromise = false
```

peterbe:
That big PR has landed now. You said this PR can wait :)

I'm really looking forward to this PR!

peterbe:
One way of testing this is:
```
node --trace-warnings bin/minimalcss.js -d --verbose http://localhost:8080/404css.html
```
where that URL is the [404css.html](https://github.com/peterbe/minimalcss/blob/master/examples/404css.html) from the examples folder. 
What happens is that [this line](https://github.com/peterbe/minimalcss/blob/9cd3f8f0716637fb77ac00653f44c06499a38ed3/src/run.js#L113) happens but the whole process isn't stop. 

Eventually you get a strange error about `TypeError: Cannot read property 'type' of undefined` because [this code](https://github.com/peterbe/minimalcss/blob/9cd3f8f0716637fb77ac00653f44c06499a38ed3/src/run.js#L253) assumes there was a successful download and parse (to AST) for every URL covered. 

So that's a solid way of testing. It should bail and crash early as soon as a CSS download fails.

peterbe:
It works great! Thank you!
This deserves a new minor release.

Issue #62, opened 2018-01-01
Alternative approach to "Evaluate DOM multiple times"


stereobooster:
Related #52

We can use chrome debug protocol CSS coverage instead, but this also will include inline CSS rules see https://github.com/peterbe/minimalcss/pull/58:

```js
  //Open Headless Chrome
  const browser = await puppeteer.launch();

  //Open a new tab
  const page = await browser.newPage();

  //Start sending raw DevTools Protocol commands are sent using `client.send()`
  //First off enable the necessary "Domains" for the DevTools commands we care about
  const client = page._client;
  await client.send('Page.enable');
  await client.send('DOM.enable');
  await client.send('CSS.enable');

  //Start tracking CSS coverage
  await client.send('CSS.startRuleUsageTracking');

  await page.goto(url,  { waitUntil: ['networkidle0'] });
  
  const { rules } = await client.send('CSS.takeCoverageDelta');
  
  const usedRules = rules.coverage
    .filter((rule) => {
      return rule.used;
    });
```

References:
- http://blog.cowchimp.com/monitoring-unused-css-by-unleashing-the-devtools-protocol/
- https://github.com/gajus/usus/blob/e6ecc83a137ca1f2317ac214c9c0d85be45475b2/src/usus.js#L274

stereobooster:
currently not possible because of bugs in chrome css coverage. See https://github.com/peterbe/minimalcss/pull/110 for details

PR #66, opened 2018-01-06
[#62] trying CSS.takeCoverageDelta

stereobooster:
It doesn't work :/ Maybe need to open issue in puppeteer tracker


stereobooster:
Update: it works as standalone example https://github.com/stereobooster/puppeteer-example

stereobooster:
will reopen when we add tests

PR #68, opened 2018-01-07
refactored page.on('request'

stereobooster:
We need to load: `document`, `javascript`, `stylesheet` and `xhr`. Everything else can be skipped (theoretically). WDYT?
  
Do not merger this yet. It seems value of resourceType depends on mime/type. So it can fail for simplest static web server

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -64,13 +64,14 @@ const postProcessKeyframes = ast => {
 
 /**
  *
- * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean }} options
+ * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, loadfonts: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean }} options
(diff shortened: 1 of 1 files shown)


peterbe:
Oh. So it doesn't work. Strangely (and stupidly) I read the code before I read your description.

stereobooster:
I thought it was working, then I tested it again and it appeared that it doesn't work, so I updated description of PR

peterbe:
I think you should break it out into its own function (perhaps in that `utils.js` file)
Instead of: 
```javascript
if (request.resourceType === 'image' && !loadimages) {
```
or (the old thing)
```javascript
if (/\.(png|jpg|jpeg|gif|webp)$/.test(request.url.split('?')[0])  && !loadimages) {
```
Make it something like
```javascript
if (utils.sniffRequestType(request, 'image') && !loadimages) {
```

It'll make it less noisy here in the main function and that util function can leverage `resourceType` if it's supported or whatever functionality it needs to do to make it work.

peterbe:
Can you try rebasing this on master and try with puppeteer 1.0.0?

stereobooster:
will close for now. Maybe will get back later

PR #83, opened 2018-01-19
Keyframe selectors should be avoided fixed 79

peterbe:
It works!

@lahmatiy Thank you for the tip! This actually works.


examples/keyframe.css
```diff
@@ -0,0 +1,10 @@
+@keyframes RotateSlot {
+  3% { margin-top: -2em }
+  from { transform: rotate(0deg)}
+}
+
+.SomeSelector {
+  font-size: 4px;
+  /* animation: RotateSlot infinite 5s linear; */
```
stereobooster: why this commented?

PR #85, opened 2018-01-21
[#4] First test

stereobooster:


tests/main.test.js
```diff
@@ -0,0 +1,58 @@
...
+test('handles JS errors', async () => {
+  expect.assertions(1)
+  try {
+    await runMinimalcss('jserror')
+  } catch (e) {
+    expect(e.message).toMatch('Error: unhandled')
+  }
+})
+
+// test('handles 404 CSS file', async () => {
```
stereobooster: this fails for some reason

tests/examples/css-relative-path.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <link href="images/css-relative-path.css" rel=stylesheet>
-    <link href=https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css rel=stylesheet>
+    <!-- <link href=https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css rel=stylesheet> -->
```
stereobooster: need to add test case for 3rd-party server and relative pathes

stereobooster:
I implemented some basic tests. Good enough for start

peterbe:
Wow wow wow! So looking forward to digging into this. I've used up my side-project time for the day but I'm really eager to land this. It looks really promising.

peterbe:
One immediate problem that jumped out at me is that the port number `:3000` is used by several of my React apps because `create-react-app` defaults to that port. Can we pick a more "unusual" port number?

stereobooster:
We can use https://www.npmjs.com/package/detect-port-alt to detect if port occupied and [even detect application](https://github.com/facebook/create-react-app/blob/19e0bb1881e24fb1ee3fe421413d4d87e67f68dd/packages/react-dev-utils/getProcessForPort.js#L27)

peterbe:
👍  on both.

PR #88, opened 2018-01-23
more tests

stereobooster:


stereobooster:
@peterbe good for review. I will add more later

stereobooster:
should we remove `examples` folder?

peterbe:
You're amazing @stereobooster !

peterbe:
> should we remove examples folder?

Yes. It's confusing that it's there if we're not actively using it. 
It would be nice to document how you can use `tests/examples` when debugging outside of running unit tests. I for one, open a separate terminal and go into the `examples` directory and run `http-server` and then test minimalcss against `localhost:8080/...`.

stereobooster:
I wanted to merge other pr and rebase this one, but erased history instead

stereobooster:
I will restore it later

PR #89, opened 2018-01-23
control waitUntil from cli

peterbe:
The diff is big. It kinda escalated. First I made it possible to control the `waitUntil` option from the cli. Then I tested it with a fat page and got weird errors. 
```sh
$ ./bin/minimalcss.js -d --waituntil load https://www.quantamagazine.org/what-makes-the-hardest-equations-in-physics-so-difficult-20180116/
TypeError: Cannot read property 'type' of undefined
    at Object.<anonymous> (/Users/peterbe/dev/JAVASCRIPT/minimalcss/node_modules/css-tree/lib/walker/create.js:12:17)
    at walkNode (/Users/peterbe/dev/JAVASCRIPT/minimalcss/node_modules/css-tree/lib/walker/create.js:155:19)
    at Object.walk (/Users/peterbe/dev/JAVASCRIPT/minimalcss/node_modules/css-tree/lib/walker/create.js:214:9)
    at allHrefs.forEach.href (/Users/peterbe/dev/JAVASCRIPT/minimalcss/src/run.js:375:13)
    at Set.forEach (native)
    at Object.minimalcss [as run] (/Users/peterbe/dev/JAVASCRIPT/minimalcss/src/run.js:372:12)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```
The assumption is that when it comes time to check every found `<link href>` we assume that they have *all* been downloaded and parsed. If you run `minimalcss` "too impatiently" that breaks. So I added a check. 
Perhaps the error message should be a bit more user-friendly and explain "You might have to set a different 'waitUntil' to give the external stylesheets more time to download."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Switching to wait for [`networkidle0`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagegotourl-options)
   instead to allow the page slightly more time to finish more XHR and static
-  resources.
+  resources. [pull#87](https://github.com/peterbe/minimalcss/pull/87)
 
 # 0.6.2
 
--- a/src/run.js
+++ b/src/run.js
@@ -129,6 +129,7 @@ const processPage = ({
     const debug = options.debug || false
     const loadimages = options.loadimages || false
     const withoutjavascript = options.withoutjavascript || false
+    const waitUntil = options.waitUntil || ['domcontentloaded', 'networkidle0']
 
     try {
       if (options.userAgent) {
@@ -255,15 +256,13 @@ const processPage = ({
       // Second, goto the page and evaluate it on the 'networkidle2' event.
       // This gives the page a chance to load any <script defer src="...">
       // and even some JS that does XHR requests right after load.
-      response = await page.goto(pageUrl, {
-        waitUntil: ['domcontentloaded', 'networkidle0']
-      })
+      response = await page.goto(pageUrl, { waitUntil })
       if (!response.ok()) {
         return safeReject(
           new Error(`${response.status()} on ${pageUrl} (second time)`)
         )
       }
-      const evalNetworkIdle = await page.evaluate(() => {
+      const evalWithJavascript = await page.evaluate(() => {
         // The reason for NOT using a Set here is that that might not be
         // supported in ES5.
         const hrefs = []
@@ -294,9 +293,9 @@ const processPage = ({
         }
       })
 
-      const htmlNetworkIdle = evalNetworkIdle.html
-      doms.push(cheerio.load(htmlNetworkIdle))
-      evalNetworkIdle.hrefs.forEach(href => {
+      const htmlWithJavascript = evalWithJavascript.html
+      doms.push(cheerio.load(htmlWithJavascript))
+      evalWithJavascript.hrefs.forEach(href => {
         allHrefs.add(href)
       })
 
@@ -308,7 +307,7 @@ const processPage = ({
 
 /**
  *
- * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any }} options
+ * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any, waitUntil: (string|string[]) }} options
  * @return Promise<{ finalCss: string, stylesheetContents: { [key: string]: string } }>
  */
 const minimalcss = async options => {
@@ -354,7 +353,15 @@ const minimalcss = async options => {
     }
   }
 
-  // All URLs have been opened, and we now have multiple DOM objects.
+  // All URLs have been opened, and we now have multiple DOM (cheerio) objects.
+  // But first check that every spotted stylesheet (by <link> tags)
+  // got downloaded.
+  const missingASTs = [...allHrefs].filter(url => !stylesheetAsts[url])
+  if (missingASTs.length) {
+    throw new Error(
+      `Found stylesheets that failed to download (${missingASTs})`
+    )
+  }
 
   // Now, let's loop over ALL links and process their ASTs compared to
   // the DOMs.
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 * Switching to wait for [`networkidle0`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagegotourl-options)
   instead to allow the page slightly more time to finish more XHR and static
-  resources.
+  resources. [pull#87](https://github.com/peterbe/minimalcss/pull/87)
```
peterbe: Unrelated. Just a fixup I had uncommitted.

src/run.js
```diff
@@ -248,15 +249,13 @@ const processPage = ({
...
-        waitUntil: ['domcontentloaded', 'networkidle0']
-      })
+      response = await page.goto(pageUrl, { waitUntil })
       if (!response.ok()) {
         return safeReject(
           new Error(`${response.status()} on ${pageUrl} (second time)`)
         )
       }
-      const evalNetworkIdle = await page.evaluate(() => {
+      const evalWithJavascript = await page.evaluate(() => {
```
peterbe: I just wanted to rename the variable now that the `waitUntil` is "more variable".

src/run.js
```diff
@@ -347,7 +346,15 @@ const minimalcss = async options => {
...
   }
 
-  // All URLs have been opened, and we now have multiple DOM objects.
+  // All URLs have been opened, and we now have multiple DOM (cheerio) objects.
+  // But first check that every spotted stylesheet (by <link> tags)
+  // got downloaded.
+  const missingASTs = [...allHrefs].filter(url => !stylesheetAsts[url])
+  if (missingASTs.length) {
+    throw new Error(
+      `Found stylesheets that failed to download (${missingASTs})`
```
peterbe: Here, making this string, there's an opportunity to better explain what *might* have failed. Is it necessary do you think?

tests/main.test.js
```diff
@@ -49,6 +48,19 @@ test('handles JS errors', async () => {
   }
 })
 
+test('handles impatient CSS downloads', async () => {
+  expect.assertions(1)
+  try {
+    await runMinimalcss('slowcss', {
+      // The most impatient option. The `slowcss.html` fixture uses
+      // a <link> href that takes a whole second to download.
+      waitUntil: 'load'
```
stereobooster: It will generate errors most of the time. Maybe forbid this option?

peterbe: Good point. You're right, it'd happen even on a page that doesn't have any slow XHR requests. Hmm... Can there be any value in ever running with `'load'`?

stereobooster: > Can there be any value in ever running with 'load'?

I can't see any

peterbe:
@stereobooster Updated.

stereobooster:
Here is my thoughts:

`react-snap` works with `networkidle0`. There is also `waitFor` option exposed, but is used after `networkidle0`.

```
await page.goto(pageUrl, { waitUntil: "networkidle0" });
if (options.waitFor) await page.waitFor(options.waitFor);
```

I use `networkidle0`, because this is the only way to make sure page is fully loaded. Otherwise there are no guaranties. My opinion we do not need to expose this config, just yet.

Do you have use case for this config?

peterbe:
> Do you have use case for this config?

Good question! I don't like that the API options *aren't* available from the cli. That's what started this PR. The default is `networkidle0` in both the core API and the cli. 

It's hard to imagine a scenario where you can accept some XHR queries to *not* finish. Even if you have some really slow loading banner ad .js (or something) (e.g. the [carbon ads on my personal blog](https://cdn.carbonads.com/carbon.js?zoneid=1673&serve=C6AILKT&placement=peterbecom) then you probably should wait because the banner ad might have important effects on how the ad is displayed. 

Perhaps it was a silly thing to make it an option in the first place. But this PR is technically not wrong. It might make the cli's `--help` output more noisy than it needs to be :)

peterbe:
This `options.waitFor` looks a lot more interesting though. However, what can you possibly do with it better if you already have set `options.waitFor` to `networkidle0`?

stereobooster:
> Good question! I don't like that the API options aren't available from the cli.

Remove it from API - that will solve inequality and make code smaller

About `options.waitFor`

I did small error in the scope of tasks for `react-snap`. The main purpose is to generate HTML snapshots, but also it can generate screenshots. When you do screenshots sometimes you need to add additional wait time to make sure all paints finished. `options.waitFor` is hacky shortcut, maybe I will remove it together with screenshot functionality

peterbe:
Apart from the git conflict on this PR, I think we should go ahead. The feature doesn't hurt and all the code is now complete. You're right in that it's nuts to NOT use `networkidle0` but you never know when it might come in handy to be able to change that. 

I'll fix the conflict and ask that we land this. Good idea?

stereobooster:
It's your call. The more feature you add, the more things you need to support (more edge cases). I strongly advice to remove it. If we will find case when we need it we can add it later

src/run.js
```diff
@@ -129,6 +129,22 @@ const processPage = ({
...
     const withoutjavascript = options.withoutjavascript || false
+    const waitUntil = options.waitUntil || ['domcontentloaded', 'networkidle0']
+
+    // 'waitUntil' can be a string or an array of strings.
+    // Whatever it is, do not allow 'load' even though it's a valid
+    // option to puppeteer's page.goto() because it's near impossible
+    // to ever figure out what the minimal CSS is if we never get a chance
+    // to download the external stylesheets.
+    if (
+      (typeof waitUntil === 'string' && waitUntil === 'load') ||
```
stereobooster:
```js
if (waitUntil !== 'networkidle0') {
  throw new Error(`networkidle0 is the only supported option at the moment`);
}
```

not trolling

peterbe:
> I strongly advice to remove it.

From the cli and from the main API?

stereobooster:
> From the cli and from the main API?

Yes from both

peterbe:
I'm so tempted to agree with you. But for now I'm going to just leave it in. The changelog has already been written and it'd feel nuts to go back on that. 

There are bigger fish to fry and clicking merge on this is easier than going back and removing the option all together. 

Let's work on other more important features. Onwards!

peterbe:
Oh no! My bad. The only mention of this option was in the README. Nothing in the changelog. I'll just rip it all out.

PR #90, opened 2018-01-23
remove collectImportantComments

peterbe:
When @lahmatiy rewrote much of the inner stuff to use `csstree` (properly instead) we stopped to depend on this utility function.


stereobooster:
nothing to add to changelog, so I simply merged it

peterbe:
Feel free to "Squash and merge" so it appears as a single commit.

PR #91, opened 2018-01-23
right jsdoc for return stylesheetContents, fixes #80

### LINKED ISSUE #80: `stylesheetContents` in promise isn't a string
[According to our doc string](https://github.com/peterbe/minimalcss/blob/a97fa21c87a47af2bab18a4e4a3f1257fc956a49/src/run.js#L42), the main `minimalcss` function returns a promise. And inside that promise is a dict. But it says that one of the keys in that dict is a string. 
`stylesheetContents` isn't a string returned in the promise. It's a dict where every key is a `.css` file URL and every value is the complete unprocesses raw CSS as a string. 

@stereobooster I guess this is where typescript might have helped if actually used in runtime :)

peterbe:


src/run.js
```diff
@@ -302,7 +302,7 @@ const processPage = ({
 /**
  *
  * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any }} options
- * @return Promise<{ finalCss: string, stylesheetContents: string }>
+ * @return Promise<{ finalCss: string, stylesheetContents: Object.<string, string> }>
```
stereobooster: To be honest this unfamiliar syntax to me. I will need to read about it

peterbe: I just skimmed http://usejsdoc.org/tags-type.html
It says "An object with string keys and number values:" for `{Object.<string, number>}`

stereobooster: The problem is that TypeScript JSDoc is not the same as "vanilla" JSDoc. [There are tone of JSDoc flavors](https://github.com/stereobooster/type-o-rama#jsdoc-flavours). This is what supported by TypeScript https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript

peterbe: oh really. But isn't this better than `string`? :)

stereobooster: I suppose this is the right syntax for it

```typescript
Promise<{ finalCss: string, stylesheetContents: { [key: string]: string } }>
```

but typescript doesn't care (I mean `yarn tsc`)

peterbe: PR updated.

peterbe:
Not going to update CHANGELOG since it's all internal.

Issue #82, opened 2018-01-19
`Error: 307 on http://www.youtube.com/embed/VcjzHMhBtf0?rel=0`


peterbe:
Steps to reproduce:
```
▶ ./bin/minimalcss.js -d -o /tmp/bootstrap.before.css http://blog.getbootstrap.com/
Error: 307 on http://www.youtube.com/embed/VcjzHMhBtf0?rel=0
    at Page.page.on.response (/Users/peterbe/dev/JAVASCRIPT/minimalcss/src/run.js:129:29)
    at emitOne (events.js:115:13)
    at Page.emit (events.js:210:7)
    at NetworkManager.Page._networkManager.on.event (/Users/peterbe/dev/JAVASCRIPT/minimalcss/node_modules/puppeteer/lib/Page.js:94:75)
    at emitOne (events.js:115:13)
    at NetworkManager.emit (events.js:210:7)
    at NetworkManager._handleRequestRedirect (/Users/peterbe/dev/JAVASCRIPT/minimalcss/node_modules/puppeteer/lib/NetworkManager.js:182:10)
    at NetworkManager._onRequestIntercepted (/Users/peterbe/dev/JAVASCRIPT/minimalcss/node_modules/puppeteer/lib/NetworkManager.js:155:14)
    at emitOne (events.js:115:13)
    at CDPSession.emit (events.js:210:7)
```

stereobooster:
Puppeteer for some reason considers only 2xx responses as successful.

```js
ok() {
  return this._status >= 200 && this._status <= 299;
}
```

https://httpstatuses.com/307 is redirect.

We can do

```js
if (response.status() >= 400) {
  return safeReject(new Error(`Response status ${response.status()} on ${pageUrl}`))
}
```

PR #95, opened 2018-01-23
[#82] fix 307 error

stereobooster:


src/run.js
```diff
@@ -176,8 +176,10 @@ const processPage = ({
       page.on('response', response => {
         const responseUrl = response.url()
         const ct = response.headers()['content-type'] || ''
-        if (!response.ok()) {
-          return safeReject(new Error(`${response.status()} on ${responseUrl}`))
+        if (response.status() >= 400) {
+          return safeReject(
+            new Error(`Response status ${response.status()} on ${responseUrl}`)
+          )
```
peterbe: This makes me nervous. I need to try to understand how this works. 

If you have a page like:
```html
<link rel="stylesheet" href="http://httpbin.org/redirect-to?url=/real.css">
```
then, *that* URL won't have a `response.text()` because it's a 30X redirect. But the `stylesheetAsts` objects needs the original URL. 
What you want is to accept it (i.e. not using `response.ok()`) but we parse it, we need to do the parsing based on the css it redirects to. 

To remedy this, I think I'd like to write a unit test to simulate it. But there's so much going on in https://github.com/peterbe/minimalcss/pull/88 that perhaps it's best to wait.

src/run.js
```diff
@@ -226,8 +228,10 @@ const processPage = ({
         // First, go to the page with JavaScript disabled.
         await page.setJavaScriptEnabled(false)
         response = await page.goto(pageUrl)
-        if (!response.ok()) {
-          return safeReject(new Error(`${response.status()} on ${pageUrl}`))
+        if (response.status() >= 400) {
+          return safeReject(
+            new Error(`Response status ${response.status()} on ${pageUrl}`)
+          )
```
peterbe: Does this work? If someone does:
```sh
$ ./bin/minimalcss.js http://httpbin.org/redirect-to?url=https://www.peterbe.com
```
I definitely want it to allow the redirect to happen, but the next line `await page.content()` needs to be that of the page it redirected to.

stereobooster:
I agree with both comments. Had the same doubts. I want to craft tests for those cases

tests/main.test.js
```diff
@@ -69,3 +77,15 @@ test('media queries print removed', async () => {
...
   expect(finalCss).toEqual(result)
 })
+
+test('handles 307 CSS file', async () => {
+  const result = 'p{color:violet}'
+  const { finalCss } = await runMinimalcss('307css')
+  expect(finalCss).toEqual(result)
+})
+
+// test('handles 307 HTML file', async () => {
```
stereobooster: https://github.com/stereobooster/puppeteer-example/tree/307-issue it works as standalone version :/

src/run.js
```diff
@@ -183,10 +183,22 @@ const processPage = ({
...
+            )
+          } else if (response.status() >= 300) {
+            page.evaluate(url => {
+              const linkTag = document.createElement('link')
+              linkTag.setAttribute('rel', 'stylesheet')
+              linkTag.setAttribute('href', url)
+              document.body.appendChild(linkTag)
+            }, response.headers().location)
+            response.text = () => Promise.resolve('')
+          }
```
peterbe: https://github.com/peterbe/minimalcss/pull/106

src/run.js
stereobooster: yes it does work without additional code, this part was causing troubles

```diff
       page.on('response', response => {
          const responseUrl = response.url()
          const ct = response.headers()['content-type'] || ''
-        if (!response.ok()) {
-          return safeReject(new Error(`${response.status()} on ${responseUrl}`))
-        }
```

stereobooster:
will close for npw

PR #101, opened 2018-01-24
remove print media at rules

peterbe:
I took the fixture you started in https://github.com/peterbe/minimalcss/pull/88 and copied it into this. Then I made the necessary code change so it actually makes the minimal CSS remove the `@media print {...}` block.

### DIFF
--- a/tests/main.test.js
+++ b/tests/main.test.js
@@ -63,3 +63,9 @@ test('handles 404 CSS file', async () => {
     expect(e.message).toMatch('404 on')
   }
 })
+
+test('media queries print removed', async () => {
+  const result = ''
+  const { finalCss } = await runMinimalcss('media-queries-print')
+  expect(finalCss).toEqual(result)
+})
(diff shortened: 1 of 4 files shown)


peterbe:
I hope this doesn't complicate things for your tests PR. If so, this can wait.

tests/main.test.js
```diff
@@ -63,3 +63,9 @@ test('handles 404 CSS file', async () => {
     expect(e.message).toMatch('404 on')
   }
 })
+
+test('media queries print removed', async () => {
```
stereobooster: I have test (in big PR) that shows that print media queries stays.

peterbe: I know and I think that's "wrong". I.e. the `@media print ...` blocks *should* be removed. 
I suggest you remove "your version" and then merging with this PR will be easier for you.

stereobooster: I mean my tests are stating facts about current code - exposing right and wrong cases, there are more wrong cases. As soon as you fix code you update tests. I just referred to the fact that there will be merge conflict nothing else

stereobooster:
this needs update in changelog

PR #106, opened 2018-01-25
crazy redirects of stylesheets, fixes #82

### LINKED ISSUE #82: `Error: 307 on http://www.youtube.com/embed/VcjzHMhBtf0?rel=0`

peterbe:
@stereobooster 
This is work in progress. I was writing a long complicated comment on https://github.com/peterbe/minimalcss/pull/95/files and decided it's better to make my point using code. 

What I didn't like was that forced injection of new stylesheets. Especially if the redirect loop is more than exactly 1. 

This PR contains a solution to redirect-of-redirects. What do you think?

There's a lot more work to do here. In particular, I need to write a unit test.


tests/examples/redirects.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <!-- <link href="http://httpbin.org/redirect-to?url=http%3A//minimalcss.local/finally.css" rel=stylesheet> -->
+    <link href="http://httpbin.org/redirect-to?url=http%3A//httpbin.org/redirect-to%3Furl%3Dhttp%253A//minimalcss.local/finally.css" rel=stylesheet>
```
peterbe: I was having problems with content-types with my `http-server` so I set up a little nginx with a local domain. 
I'm obviously going to rewrite this using `fastify` for the tests.

stereobooster:
> There's a lot more work to do here. In particular, I need to write a unit test.

take unit tests from #95

peterbe:
This is what you see in Chrome:
screen shot 2018-01-25 at 4 29 29 pm
(I don't know what that `page.bundle.js` is)

peterbe:
And the output works as expected:
```sh
▶ ./bin/minimalcss.js -d http://minimalcss.local/redirects.html
p{color:pink}
```

stereobooster:
@peterbe sorry pushed directly to your branch, wanted to push to my branch. But I added tests

stereobooster:
Second day in a row I do not make friends with git :/

tests/main.test.js
```diff
@@ -69,3 +77,15 @@ test('media queries print removed', async () => {
...
   expect(finalCss).toEqual(result)
 })
+
+test('handles 307 CSS file', async () => {
+  const result = 'p{color:violet}'
+  const { finalCss } = await runMinimalcss('307css')
+  expect(finalCss).toEqual(result)
+})
+
+// test('handles 307 HTML file', async () => {
```
stereobooster: you can uncomment this, not sure why it was commented

src/run.js
```diff
@@ -182,10 +183,14 @@ const processPage = ({
       // To build up a map of all downloaded CSS
       page.on('response', response => {
         const responseUrl = response.url()
+
         const ct = response.headers()['content-type'] || ''
-        if (!response.ok()) {
+        if (response.status() >= 400) {
```
stereobooster: thing is that even html redirects go through this logic. You can simply move this block below `if (ct.indexOf('text/css') > -1 || /\.css$/i.test(responseUrl)) {` and it will work for html redirect case without additional logic. But it doesn't work for css redirects case

peterbe: I'm not entirely sure I understood your comment but I poked around and noticed that the problem was that the `Location` header had a relative URL. The lines below, quite nicely, turns it into an absolute URL. That solved the breaking tests.

stereobooster:
```diff
-        if (!response.ok()) {
-          return safeReject(new Error(`${response.status()} on ${responseUrl}`))
-        }
         if (ct.indexOf('text/css') > -1 || /\.css$/i.test(responseUrl)) {
+          if (!response.ok()) {
+            return safeReject(new Error(`${response.status()} on ${pageUrl}`))
+          }
```

https://github.com/peterbe/minimalcss/pull/107/files#diff-f712f2169bd3dca93334576e35acff9fL186

peterbe:
@stereobooster When you review, can you leave an approve comment and I'll do the merge, please.

src/run.js
stereobooster: move all logic about `response.status()` below `if (ct.indexOf('text/css') > -1 || /\.css$/i.test(responseUrl)) {`. This logic should be applied only for CSS documents, nothing else

peterbe: I think I understand what you mean. However, it sounds risky. 
What if I have: `<link rel="stylesheet" href="http://example.com/redirect-to-the-real-css.cgi">`. 
When this resource is coming through our `page.on('response'` handler, its filename doesn't end with `.css` and it's content type is not going to be `text/css`.

For example:
```sh
▶ curl -v http://cdn-2916.kxcdn.com/static/css/base.min.7d48dc16ca51.css\?cachebust\=1 > /dev/null
[snip]
> GET /static/css/base.min.7d48dc16ca51.css?cachebust=1 HTTP/1.1
> Host: cdn-2916.kxcdn.com
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 301 Moved Permanently
< Server: keycdn-engine
< Date: Fri, 26 Jan 2018 15:24:10 GMT
< Content-Type: text/html
< Content-Length: 178
< Connection: keep-alive
< Location: https://cdn-2916.kxcdn.com/static/css/base.min.7d48dc16ca51.css?cachebust=1
< X-Edge-Location: usch
< Access-Control-Allow-Origin: *
<
```
If I only wrote down where this redirect **if** it's `.css` or `text/css` content-type, then this would fail. 

By writing down every single redirect, we're safer.

stereobooster: We can add test for this case

peterbe:
A lot happened. It kinda escalated into less-than-related other bits and pieces. 

So now, the HTML as a `<link>` tag that doesn't look like a `.css` and all things work. 

I also noticed that we don't have a test case for when a HTML document contains a `<link>` that should be ignored. I don't remember the original motivation for that but now we have a test.

src/run.js
```diff
@@ -152,15 +154,16 @@ const processPage = ({
 
       await page.setRequestInterception(true)
       page.on('request', request => {
-        if (/data:image\//.test(request.url())) {
+        const requestUrl = request.url()
```
peterbe: I just noticed we used `request.url()` so many times that it felt right to make a local variable.

src/run.js
peterbe: Tested for in the latest push.

src/run.js
```diff
@@ -377,6 +395,12 @@ const minimalcss = async options => {
     })
   }
   allHrefs.forEach(href => {
+    while (redirectResponses[href]) {
+      href = redirectResponses[href]
+    }
+    if (skippedUrls.has(href)) {
```
peterbe: This lacks a comment. I'll add one.

tests/examples/redirected.css
```diff
@@ -0,0 +1 @@
+p { color: violet }
```
peterbe: I just renamed `redirect.css` to `redirected.css` because the name `redirect.css` sounds like it's *going to* redirect, when in fact it's more like the "destination".

peterbe:
@stereobooster You approved this before but quite a lot has changed. Sorry. Can you take another look?

src/run.js
```diff
@@ -152,15 +154,16 @@ const processPage = ({
...
       page.on('request', request => {
-        if (/data:image\//.test(request.url())) {
+        const requestUrl = request.url()
+        if (/data:image\//.test(requestUrl)) {
           // don't need to download those
           request.abort()
         } else if (
           !loadimages &&
-          /\.(png|jpg|jpeg|gif|webp)$/.test(request.url().split('?')[0])
+          /\.(png|jpg|jpeg|gif|webp)$/.test(requestUrl.split('?')[0])
```
stereobooster: for later - add `svg`

tests/main.test.js
```diff
@@ -124,3 +133,24 @@ test('form elements', async () => {
   expect(finalCss).toMatch('input[type=checkbox]:checked')
   expect(finalCss).toMatch('option:selected')
 })
+
+test('handles 307 CSS file', async () => {
+  const result = 'p{color:violet}'
+  const { finalCss } = await runMinimalcss('307css')
+  expect(finalCss).toEqual(result)
```
stereobooster:
```js
expect(finalCss).toEqual('p{color:violet}')
```

I can do this later

tests/main.test.js
```diff
@@ -124,3 +133,24 @@ test('form elements', async () => {
...
+})
+
+test('handles 307 HTML file', async () => {
+  const result = 'p{color:violet}'
+  const { finalCss } = await runMinimalcss('307')
+  expect(finalCss).toEqual(result)
```
stereobooster:
```js
expect(finalCss).toEqual('p{color:violet}')
```

PR #107, opened 2018-01-26
[#82] fix 307 error

stereobooster:
Tests and minimal implementation

### DIFF
--- a/tests/main.test.js
+++ b/tests/main.test.js
@@ -7,6 +7,14 @@ fastify.register(require('fastify-static'), {
   root: path.join(__dirname, 'examples')
 })
 
+fastify.get('/307.css', (req, reply) => {
+  reply.redirect(307, '/redirect.css')
+})
+
+fastify.get('/307.html', (req, reply) => {
+  reply.redirect(307, '/redirect.html')
+})
+
 let browser
 
 const runMinimalcss = path => {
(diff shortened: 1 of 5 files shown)


tests/main.test.js
```diff
@@ -134,3 +142,15 @@ test('form elements', async () => {
...
   expect(finalCss).toEqual(result)
 })
+
+test('handles 307 HTML file', async () => {
+  const result = 'p{color:violet}'
+  const { finalCss } = await runMinimalcss('307')
+  expect(finalCss).toEqual(result)
+})
+
+test.skip('handles 307 CSS file', async () => {
```
stereobooster: to make this test work we need

```js

          if (response.status() >= 400) {
            return safeReject(
              new Error(
                `Response status ${response.status()} on ${responseUrl}`
              )
            )
          } else if (response.status() >= 300) {
            page.evaluate(url => {
              const linkTag = document.createElement('link')
              linkTag.setAttribute('rel', 'stylesheet')
              linkTag.setAttribute('href', url)
              document.body.appendChild(linkTag)
            }, response.headers().location)
            response.text = () => Promise.resolve('')
          }
```

peterbe:
Let's use https://github.com/peterbe/minimalcss/pull/106 instead.

PR #110, opened 2018-01-26
[#62] trying CSS.takeCoverageDelta

stereobooster:
Not all tests pass, because there are issues in chrome CSS coverage. I just want to show you that this is possible. As soon as they will fix chrome bugs we can simplify `minimalcss` a lot.

Take a look and close it.

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -5,7 +5,6 @@ const puppeteer = require('puppeteer')
 const csso = require('csso')
 // @ts-ignore
 const csstree = require('css-tree')
-const cheerio = require('cheerio')
 const utils = require('./utils')
 const url = require('url')
 
@@ -109,9 +108,7 @@ const processPage = ({
   options,
   pageUrl,
   stylesheetAsts,
-  stylesheetContents,
-  doms,
-  allHrefs
+  stylesheetContents
 }) =>
   new Promise(async (resolve, reject) => {
     // If anything goes wrong, for example a `pageerror` event or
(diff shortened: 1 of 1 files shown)


peterbe:
I didn't read it super carefully but I saw you commented out the page load with javascript disabled. 
Is that deliberate or is not going to be possible if you use the devtools?

I'm quite excited about the idea of it giving a minimal CSS based on ALL CSS loaded into memory and not just the external CSS.

peterbe:
So what's the Chrome bug to pin this to?

stereobooster:
> I didn't read it super carefully but I saw you commented out the page load with javascript disabled.
Is that deliberate or is not going to be possible if you use the devtools?

By idea - it is not needed. Coverage should take into account every used CSS rule from my POV. But this is not the case right now.

> I'm quite excited about the idea of it giving a minimal CSS based on ALL CSS loaded into memory and not just the external CSS.

But I filtered it out

```js
const { header } = stylesheet
if (
  header.isInline ||
  header.sourceURL === '' ||
  header.sourceURL.startsWith('blob:')
) {
  inlineStylesheetIndex.add(header.styleSheetId)
}
```

stereobooster:
> So what's the Chrome bug to pin this to?

- https://bugs.chromium.org/p/chromium/issues/detail?id=765088
- https://bugs.chromium.org/p/chromium/issues/detail?id=744202&q=css%20Coverage%20&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified
- https://bugs.chromium.org/p/chromium/issues/detail?id=734470&q=css%20Coverage%20&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified
- https://bugs.chromium.org/p/chromium/issues/detail?id=717195&q=css%20Coverage%20&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified

stereobooster:
Will close for now as not actionable

Issue #117, opened 2018-01-27
Add tests for CSS variables


stereobooster:
```css
element {
  --main-bg-color: brown;
}
```

https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables

stereobooster:
Also add test for vendor specific css

```css
-webkit-transition: all 4s ease;
-moz-transition: all 4s ease;
-ms-transition: all 4s ease;
-o-transition: all 4s ease;
transition: all 4s ease;
```

stereobooster:
#119

PR #119, opened 2018-01-27
#117 more tests

stereobooster:


tests/examples/css-variables.css
```diff
@@ -0,0 +1,8 @@
+:root {
```
stereobooster: this is what they use in bootstrap

tests/main.test.js
```diff
@@ -152,3 +152,23 @@ test("deliberately skipped .css shouldn't error", async () => {
...
 })
+
+test.skip('leaves used css variables', async () => {
+  const { finalCss } = await runMinimalcss('css-variables')
+  expect(finalCss).toMatch('--main-bg-color:')
+})
+
+test.skip('removes unused css variables', async () => {
+  const { finalCss } = await runMinimalcss('css-variables')
+  expect(finalCss).not.toMatch('--unused-color:')
```
peterbe: I think these two above tests test almost the same thing and thus they can be in the same test. 
E.g.
```javascript
test.skip('css variables', async () => {
  expect.assertions(2)
  const { finalCss } = await runMinimalcss('css-variables')
  expect(finalCss).toMatch('--main-bg-color:')
  expect(finalCss).not.toMatch('--unused-color:')
})

File: tests/main.test.js [hunk 3]
```diff
@@ -152,3 +152,23 @@ test("deliberately skipped .css shouldn't error", async () => {
   })
   expect(finalCss).toEqual('p{color:brown}')
 })
+
+test.skip('leaves used css variables', async () => {
```
peterbe: Can you add a comment above this one *why* this test is currently skipped. 
Perhaps a comment that contains a URL to a filed issue about it.

tests/examples/css-variables.html
```diff
@@ -0,0 +1,10 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <link href="css-variables.css" rel=stylesheet>
+  </head>
+  <body>
+    <p>Only a P tag in this document. And images too.</p>
```
peterbe: There are no images in this HTML.

tests/examples/css-variables.css
peterbe: Perhaps that should be in a CSS comment right before. So we don't forget.

peterbe:
@stereobooster wanna rebase and land this?

peterbe:
@stereobooster Feel free to squash and merge.

PR #123, opened 2018-01-29
test that order matters

peterbe:
To make this work, nothing had to be done to the code. But it feels like such an important feature that I didn't want it to go without a good test. 

It basically confirms that:
```css
p {
  font-size: 10px;
  color: violet;
}
```
**∪**
```css
p {
  font-size: 16px;
  font-style: italic;
}
```
**=**
```css
p {
  color: violet;
  font-size: 16px;
  font-style: italic;
}
```

This is what you get when you open `localhost:8080/inheritance.html` in the browser too.

### DIFF
--- a/tests/examples/inheritance-first.css
+++ b/tests/examples/inheritance-first.css
@@ -1,0 +1,4 @@
+p {
+  font-size: 10px;
+  color: violet;
+}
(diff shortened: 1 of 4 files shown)


stereobooster:
```
    Expected value to equal:
      "p{color:violet;font-size:16px;font-style:italic}"
    Received:
      "p{font-style:italic;font-size:10px;color:violet}"
```

peterbe:
What? How?

peterbe:
Added a new issue: https://github.com/peterbe/minimalcss/issues/130

This PR is probably "as good as it gets" the tests isn't bad. Just sad that the code doesn't always work.

PR #128, opened 2018-02-03
end-to-end tests

peterbe:
crappy start but all I had time to do. At least it tests the cli a little bit.


peterbe:
This PR now depends on this: https://github.com/peterbe/minimalcss/pull/133

bin/e2e.js
```diff
@@ -0,0 +1,73 @@
...
+  const opened = spawn('./bin/minimalcss.js', options)
+  assertStatus(opened)
+  const t1 = new Date()
+  const t = (t1 - t0) / 1000
+  console.log(`Took ${t.toFixed(2)}s`)
+  return Promise.resolve(opened)
+}
+
+// Simplest form of opening
+openUrl('https://en.wikipedia.org/wiki/List_of_HTTP_status_codes').then(
```
stereobooster: usage of real network decreases stability of test. It can fail due to external conditions, not because code is wrong

peterbe: Indeed and that's unfortunate but it's a really useful test. 
One idea is to make it something like this:

```javascript
if (failed) {

    if (inCI) {
        console.warning(`WARNING! The ${url} end-to-end test failed: ${errorMessage}`)
        console.log("End-2-end breakage doesn't stop the CI build but it's a bit sad")
        process.exit(0)
    } else {
        process.exit(1)
    }

```

peterbe: Hmm... I'm not so sure what to do about this. If I don't let it error, breaking the cli app will not be noticed. An alternative is to write unit tests for the cli app by mocking the the system arguments or something. But that's a bunch of worktime which I don't have. 

I might just land this as is and break it up in the future if it turns out to be too many misleading build failures due to unfortunate luck and moving targets.

PR #131, opened 2018-02-04
reduce by correct order of hrefs, fixes #130

### LINKED ISSUE #130: Order sometimes not respected
In [this PR](https://github.com/peterbe/minimalcss/pull/123) we added a unit test that asserts that the order of the stylesheets is respected. The tests passed but sometimes it fails. 
[Here's one such example](https://travis-ci.org/peterbe/minimalcss/builds/336805131?utm_source=github_status&utm_medium=notification). This is an unrelated Travis build. 

@stereobooster also [pointed it out here](https://github.com/peterbe/minimalcss/pull/123#issuecomment-362392241).
Need to try to understand why this is unpredictable.

peterbe:


peterbe:
It might be possible to do with a `Map` instead. Going to try that first.

peterbe:
Nah. I tried using a `Map`. It got a lot more complicated than this version.

peterbe:
Of course I can't use a `Map`. It's not about the order items are inserted into `stylesheetAsts` since that order is dependent on when responses finish. Which is uninteresting. What's interesting is only the order with which the `<link>` tags appear in the HTML.

PR #133, opened 2018-02-05
launch puppeteer args, fixes #132

### LINKED ISSUE #132: Ability to set no-sandbox
When running inside a docker container, puppeteer might fail with this message:
```
STDERR Error: Failed to launch chrome!
[0205/134736.600483:ERROR:zygote_host_impl_linux.cc(90)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
```

[react-snap has this tip](https://github.com/stereobooster/react-snap/blob/master/Readme.md#containers-and-other-restricted-environments).

I think it would be nice to be able to tell `minimalcss` to set those options. 

It would also be nice to understand what the disadvantage of those are. I.e. if using `puppeteer.launch({args: ["--no-sandbox", "--disable-setuid-sandbox"]})` makes things work in virtualized containers, why not always run like that?

peterbe:


stereobooster:
In react-snap there is section on this subject https://github.com/stereobooster/react-snap#containers-and-other-restricted-environments

Those are all puppeteer settings in `react-snap`

```
  puppeteerArgs: [],
  puppeteerExecutablePath: undefined,
  puppeteerIgnoreHTTPSErrors: false,
```

Issue #162, opened 2018-03-16
Error: 304


stereobooster:
Error trace for `minimalcss@0.6.3`

```
🔥  / Error: 304 on http://localhost:45678/
 at Promise (/node_modules/minimalcss/src/run.js:249:29)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
(node:78612) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Protocol error (Network.getResponseBody): Target closed.
(node:78612) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:78612) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: Protocol error (Network.getResponseBody): Target closed.
error An unexpected error occurred: "Command failed.
```

Update `minimalcss` to `0.7.3` error happens at  `node_modules/minimalcss/src/run.js:245:29`

`304` stands for `Not Modified` so maybe we do not need to throw error here?

stereobooster:
Update `minimalcss` to `0.7.3` error happens at  `node_modules/minimalcss/src/run.js:245:29`

stereobooster:
Shouldn't it be `if (withoutjavascript)`?

```js
      if (!withoutjavascript) {
        // First, go to the page with JavaScript disabled.
        await page.setJavaScriptEnabled(false);
        response = await page.goto(pageUrl);
        if (!response.ok()) {
          return safeReject(new Error(`${response.status()} on ${pageUrl}`));
        }
        const htmlVanilla = await page.content();
        doms.push(cheerio.load(htmlVanilla));
        await page.setJavaScriptEnabled(true);
      }
```

stereobooster:
Another finding: [Google web fonts uses User Agent Sniffing](https://stackoverflow.com/questions/19467122/how-does-google-web-fonts-decide-which-version-of-a-font-to-provide-to-my-browse), so Google fonts css should be left untouched.

peterbe:
The whole point of us looking at statuses where the code was [300, 400] is so that we can write down what the URL redirects to. That way we can "track" what Chrome does already but have a record of all redirects. 
So ultimately we're only really interested in "Was there a redirect?". We currently have this:

```javascript
        } else if (response.status() >= 300) {
          // If the 'Location' header points to a relative URL,
[...]

PR #165, opened 2018-03-16
Fix HTTP statuses

stereobooster:
Fix for https://github.com/peterbe/minimalcss/issues/162

WDYT?

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -9,6 +9,15 @@ const cheerio = require('cheerio');
 const utils = require('./utils');
 const url = require('url');
 
+const isOk = status => status === 200 || status === 304;
+
+const isRedirect = status =>
+  status === 301 ||
+  status === 302 ||
+  status === 303 ||
+  status === 307 ||
+  status === 308;
+
 /**
  * Take in a csstree AST, mutate it and return a csstree AST.
  * The mutation is about:
@@ -185,7 +194,7 @@ const processPage = ({
           return safeReject(
             new Error(`${response.status()} on ${responseUrl}`)
           );
-        } else if (response.status() >= 300) {
+        } else if (isRedirect(response.status())) {
           // If the 'Location' header points to a relative URL,
           // convert it to an absolute URL.
           // If it already was an absolute URL, it stays like that.
@@ -241,7 +250,7 @@ const processPage = ({
         // First, go to the page with JavaScript disabled.
         await page.setJavaScriptEnabled(false);
         response = await page.goto(pageUrl);
-        if (!response.ok()) {
+        if (!isOk(response.status())) {
           return safeReject(new Error(`${response.status()} on ${pageUrl}`));
         }
         const htmlVanilla = await page.content();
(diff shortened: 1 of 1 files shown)


src/run.js
```diff
@@ -9,6 +9,15 @@ const cheerio = require('cheerio');
 const utils = require('./utils');
 const url = require('url');
 
+const isOk = status => status === 200 || status === 304;
```
peterbe: We should probably stick closer to how [puppeteer does it](https://github.com/GoogleChrome/puppeteer/blob/38c6873fbbc469fe422977131bf2da8e75ec9228/lib/NetworkManager.js#L554)

Perhaps, instead of our own `isOk` function, we do something like:
```javascript
if (!(response.ok() || response.status() === 304)) {
```

Kinda strange the puppeteer doesn't consider 304 to be "OK". 

Also, we could do something like 
```javascript
if (response.status() >= 300 && response.headers().location) {
   ... it wants to redirect...
```

stereobooster:
> It hurts my soul to see such a specific list of specific error codes

but they are specified by standard

```
300 Multiple Choices
301 Moved Permanently
302 Found
303 See Other
304 Not Modified
305 Use Proxy
307 Temporary Redirect
308 Permanent Redirect
```

I picked those that make sense

src/run.js
stereobooster: note: this is how ok function looks like

```js
  ok() {
    return this._status === 0 || (this._status >= 200 && this._status <= 299);
  }
```

peterbe:
Sorry, I should have put [this comment](https://github.com/peterbe/minimalcss/issues/162#issuecomment-374370297) here in the PR instead. My bad. But still, what do you think about that idea?

stereobooster:
updated code

src/run.js
```diff
@@ -185,7 +187,7 @@ const processPage = ({
           return safeReject(
             new Error(`${response.status()} on ${responseUrl}`)
           );
-        } else if (response.status() >= 300) {
+        } else if (response.status() >= 300 && response.status() !== 304) {
```
peterbe: Can you move this second if statement part *inside*. So it's like this:
```javascript
} else if (response.status() >= 300) {
  // 304 Not Modified is an exception to the 3xx responses.
  // It's expected to not be a redirect and not have content. 
  if (response.status() !== 304) {
    ...existing code...
  }
}
```
Otherwise, you'll end up in the `} else if (resourceType === 'stylesheet') {` around line 199. 

Look at this screenshot from Firefox's Dev tools. 
screen shot 2018-03-20 at 7 34 49 am
It's smart enough to figure out that the response type is "stylesheet" but if that was to happen in `minimalcss` it would fail unpredictably on the line `response.text().then(text => {`.
Basically for 304 Not Modified we don't want to write down their redirects *or* try to read their content.

stereobooster: But for 304 we should read content. Server didn't send content - yes, but browser should have this content, so `response.text()` should work. Unless I'm confusing something - we should treat 304 as 200

peterbe: A HA! That makes sense. You're right. Arguably we should have the content already if the request was made such that the server responded with a 304. But perhaps if you pass in a `puppeteer` browser instance that already has some stuff cached, it could happen. 
I rest my case.

PR #191, opened 2018-05-01
Lerna

stereobooster:
- I do not use docker. I'm not sure what is the value of Docker in this case
- Not sure about package names: `minimalcss-lib` or `minimalcss-core`?
- As of now using local path instead of version for `minimalcss-lib`. We need to publish `minimalcss-lib` and use version instead of path and later publish `minimalcss`

`minimalcss` as of now backward compatible.


peterbe:
I haven't actually started reviewing it yet but from your description, a thought that jumps out is; Why can't it be `minimalcss-core` and `minimalcss-cli` because that's really what you want to split up isn't it?

I'm actually tempted to break out `minimalcss-cli` as its own standalone git repo and npmjs package. That's how [csso](https://github.com/css/csso) [csso-cli](https://github.com/css/csso-cli) does. 

The cli is, to me, more and more irrelevant. The way *you* use `minimalcss` in `react-snap` is to do all things `minimalcss` *except* the cli part. And so do I in https://github.com/peterbe/minimalcss-server

I know this news/opinion might feel frustrating considering this PR you've made but I've changed my personal stance with regards to the cli a bit. `minimalcss` is complex enough as it is.

peterbe:
The *advantages* of ripping out the cli into its own GitHub repo:

* Less complexity tooling. No need for fancy tools.
* Ability to focus on a core `minimalcss` features (issue/PR) without having to worry about the cli (because you might not use it)
* Less stuff in your `node_modules` if you don't need the cli if you only depend on the pure lib `minimalcss`. 
* Potentially fewer and faster tests suites.
* There's already a Node web server project called [`minimalcss-server`](https://github.com/peterbe/minimalcss-server). It doesn't yet have any tests or infra or a good README but it sets a certain fashion.
* If other people want to start tools around `minimalcss` (e.g. `minimalcss-google-firebase` (made up)) then they don't feel pressured to get it into the monorepo.

And the *disadvantages*:

* The cli might slowly drift away from the functionality of the core `minimalcss` lib.
* Hard to launch a new feature from the perspective of the cli as you need to manually track it in the core `minimalcss` lib. 
* Yet another `.travis.yml` to get right. 
* More npmjs sit-ups (even with Lerna you have to publish packages separately but at least it's tooled)
* Confusion with names. E.g. "I ran 'npm i minimalcss' now how the heck do I run it?! Argh!"
* Raises the question; If the CLI is part of the core mono-repo, why isn't `minimalcss-server` there too? What about `angular-minimalcss` or `minimalcss-webpagetest-analyzer-goof`? 
* Does this mean the core lib needs be its own organization, like https://github.com/css?

peterbe:
With that said, right now I feel the best course of action is to extract the cli and make it its own new repo and npmjs package. 

What do you think @stereobooster ?

PS. There appears to be some other juicy bits in this PR we should benefit from such as the removal of the docker stuff in CI.

stereobooster:
Monorepos are good to maintain projects which change simultaneously. If you're ok that CLI can deviate from core lib, then it is ok. Create empty `minimalcss-cli` with readme and grant me access.

> Does this mean the core lib needs be its own organization, like

We can have organisation...

peterbe:
I'm going to start a new repo just for the cli. 

@sindresorhus is leading [by](https://github.com/sindresorhus/slugify) [example](https://github.com/sindresorhus/slugify-cli).

stereobooster:
https://github.com/peterbe/minimalcss/pull/200

Issue #112, opened 2018-01-26
Would be nice to know *what* timed out


peterbe:
See 
```sh
▶ ./bin/minimalcss.js -d --verbose -o /tmp/songsearch.css https://songsear.ch/song/Starchild/Wolfgun/537590 https://songsear.ch/ https://songsear.ch/q/searching
Error: Navigation Timeout Exceeded: 30000ms exceeded
    at Promise.then (/Users/peterbe/dev/JAVASCRIPT/minimalcss/node_modules/puppeteer/lib/NavigatorWatcher.js:71:21)
    at <anonymous>
```
So most likely some asset or URL in that got stuck and took more than 30s. But I have no insight into which one it might be.

peterbe:
Curiously, this fails every time:
```sh
▶ ./bin/minimalcss.js -d --verbose -o /tmp/songsearch.css https://songsear.ch/
```

It's strange because I can see that the server did respond with a 200 OK twice. As expected!
Nginx logs:
```
73.180.105.76 - - [26/Jan/2018:16:08:29 +0000] "GET / HTTP/1.1" 200 5063 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_2) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/65.0.3312.0 Safari/537.36" "-" "songsear.ch" sn="songsear.ch" rt=0.005 ua="127.0.0.1:3000" us="200" ut="0.005" ul="14041" cs=-
73.180.105.76 - - [26/Jan/2018:16:08:30 +0000] "GET / HTTP/1.1" 200 5075 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_2) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/65.0.3312.0 Safari/537.36" "-" "songsear.ch" sn="songsear.ch" rt=0.007 ua="127.0.0.1:3000" us="200" ut="0.007" ul="14059" cs=-
```
I grepped the access logs and could see my queries coming in and finishing just fine.

peterbe:
I wrapped the `page.goto` call in a try/catch and that does catch something:
```diff
-      response = await page.goto(pageUrl, {
-        waitUntil: ['domcontentloaded', 'networkidle0']
-      })
+      try {
+        response = await page.goto(pageUrl, {
+          waitUntil: ['domcontentloaded', 'networkidle0']
+        })
+      } catch(ex) {
[...]

PR #199, opened 2018-05-13
[#112] List what timed out

stereobooster:


peterbe:
Looking forward to find the time to dig into this. Looks exciting!
Sorry for being swamped.

src/run.js
```diff
@@ -124,8 +125,20 @@ const processPage = ({
...
+    const tracker = createTracker(page);
     const safeReject = error => {
       if (!fulfilledPromise) {
+        if (error.message.startsWith('Navigation Timeout Exceeded')) {
+          const urls = tracker.urls();
+          if (urls.length > 1) {
+            error.message += `\nFor one of the following urls: ${urls.join(
+              ', '
+            )}`;
+          } else if (urls.length > 0) {
```
peterbe: This could just be `} else if (urls.length) {`

src/run.js
```diff
@@ -124,8 +125,20 @@ const processPage = ({
     // a second time.
     let fulfilledPromise = false;
 
+    const tracker = createTracker(page);
     const safeReject = error => {
       if (!fulfilledPromise) {
+        if (error.message.startsWith('Navigation Timeout Exceeded')) {
+          const urls = tracker.urls();
+          if (urls.length > 1) {
+            error.message += `\nFor one of the following urls: ${urls.join(
```
peterbe: That sentence is odd. It'd make sense if it started with `...for once of the`. 

I'm thinking we can be more direct and blunt. E.g.:
```javascript
if (url.length) {
  error.message += `Tracked URLs that have NOT finished: ${urls.join(',')}`
}
```

src/tracker.js
```diff
@@ -0,0 +1,18 @@
+const createTracker = page => {
```
peterbe: This could do with a doc string. I don't know what the right JSDoc notation should be but we ought to explain its purpose in life. 

*"A function that sets up request events on a Puppeteer.Page instance so that we can keep track of which URLs have started being request and which have finished. Ultimately its purpose is it be able to, at any point (e.g. an unexpected navigation error), be able to say which URLs have *not* finished. Having this list of URLs can be helpful when debugging why you get unpredictable timeouts."*

src/tracker.js
```diff
@@ -0,0 +1,18 @@
...
+  const requests = new Set();
```
peterbe: Why a set? How is a request hashed to be unique? 
Also what happens if you have something like:
```javascript
$(function() {
  window.setInterval(function() {
    fetch('/api/check-for-new-content').then(...)
  }, 1000)
}
```
Or if there's a repeated `` and the URL it resolves to doesn't have cache-control.

tests/main.test.js
```diff
@@ -216,3 +224,25 @@ test('accept CSSO options', async () => {
...
+});
+
+test('timeout error for resources', async () => {
+  expect.assertions(2);
+  try {
+    await runMinimalcss('with-timeout', { timeout: 2000 });
+  } catch (e) {
+    expect(e.message).toMatch('Navigation Timeout Exceeded: 2000ms exceeded');
+    expect(e.message).toMatch(
+      'For one of the following urls: http://localhost:3000/timeout.css?1, http://localhost:3000/timeout.css?2'
```
peterbe: That comma is a bit weird. Perhaps we should do it like this:
```javascript
error.message += `For one of the following urls:\n${urls.join('\n')}` 
```

peterbe:
I put a couple of more `console.log` inside `tracker.js`'s `onStarted` and `onFinished` and then analyzed the output and concluded this:

```
https://songsear.ch/song/Starchild/Wolfgun/537590                                2 ['ADDING', 'DELETING']
https://songsear.ch/static/css/main.3196ddc2.css                                 2 ['ADDING', 'DELETING']
https://songsear.ch/static/ga.js                                                 2 ['ADDING', 'DELETING']
https://songsear.ch/static/js/main.824bb95d.js                                   2 ['ADDING', 'DELETING']
https://songsearch-2916.kxcdn.com/static/albums/2017/05/12/04/63881_300x300.jpg  2 ['ADDING', 'DELETING']
https://www.google-analytics.com/analytics.js                                    2 ['ADDING', 'DELETING']
https://www.google-analytics.com/r/collect                                       2 ['ADDING', 'DELETING']
https://songsear.ch/api/song/537590/amazon                                       2 ['ADDING', 'DELETING']
https://songsear.ch/api/song/537590/youtube                                      2 ['ADDING', 'DELETING']
https://songsear.ch/api/song/537590/affiliate/amazon                             2 ['ADDING', 'DELETING']
https://songsear.ch/api/song/537590/affiliate/itunes                             2 ['ADDING', 'DELETING']
https://images-na.ssl-images-amazon.com/images/I/21hPdSeQlAL._SL75_.jpg          2 ['ADDING', 'DELETING']
https://images-na.ssl-images-amazon.com/images/I/51un9j%2BsvaL._SL75_.jpg        2 ['ADDING', 'DELETING']
https://i.ytimg.com/vi/10JQ6XzqxuU/hqdefault.jpg                                 2 ['ADDING', 'DELETING']
https://songsear.ch/api/song/537590/related                                      2 ['ADDING', 'DELETING']
```

Note that that for every URL started, there's a deleted one.

peterbe:
Basically, I think puppeteer is failing us here in that something times out that was never sent to the event system. Perhaps a service-worker or something like that. (e.g. the URLs I tested on uses service workers but beyond that there are no buggy banner ads, web workers, web sockets or anything fancy like that)

**Last but not least**, I did mess with the timeout and when I set it to 500ms it totally works. The URLs that didn't make it in 500ms are listed nicely in the error message. E.g.
```
Error: Navigation Timeout Exceeded: 500ms exceeded
For one of the following urls: https://songsear.ch/static/js/main.824bb95d.js, https://www.google-analytics.com/analytics.js
    at Promise.then (/Users/peterbe/dev/JAVASCRIPT/minimalcss/node_modules/puppeteer/lib/NavigatorWatcher.js:71:21)
    at <anonymous>
```

stereobooster:
If we can not catch it with given script then we should reopen ticket in puppeteer repo

peterbe:
> If we can not catch it with given script then we should reopen ticket in puppeteer repo

Ha! That's what I was looking to do but got distracted. I'll try to do that. 

By the way, I can run those URLs that I used to reproduce it locally. It's different in that I can't use HTTPS locally. But it's the same functionality inside the pages. Also, the other obvious difference is that when I use `songsearch.local` instead it resolves to 127.0.0.1. 
When I do this, and hawkishly watch the local Nginx logs, it always works. 
So, service workers are automatically disabled when you don't have HTTPS. That makes me, even more, believe it's the service workers that are failing.

stereobooster:
You can use self-signed certificate for HTTPS and pass flag to puppeteer to allow untested certificates

peterbe:
https://github.com/GoogleChrome/puppeteer/issues/1908#issuecomment-390214976

peterbe:
I wonder if there's a flag to tell puppeteer to tell Chrome to not bother with service-workers (because, we're not ever coming back with a hot cache).

src/run.js
peterbe: Also, I thin the word should be `URLs` (not `urls`) since it's an acronym of Uniform Resource Locator.

peterbe:
Can you rebase this up so it uses the latest puppeteer. I really think the [upgrade to puppeteer 1.4.0](https://github.com/peterbe/minimalcss/pull/214) solves the mysterious problem with strange timeout exceeded errors. At least I hope so.

stereobooster:
> I wonder if there's a flag to tell puppeteer to tell Chrome to not bother with service-workers (because, we're not ever coming back with a hot cache).

Yes. https://github.com/GoogleChrome/puppeteer/issues/1396#issuecomment-344848156

> Can you rebase this up so it uses the latest puppeteer.

Done

src/tracker.js
peterbe: Nevermind this. A set is fine. It keeps it simple. We can revisit this later.

peterbe:
Sorry. I must have forgot to run the tests after I changed the wording in the error message. Will fix that.

peterbe:
There https://github.com/stereobooster/minimalcss/pull/3

stereobooster:
I pushed this branch into this repo, you can open PR based on it, if you want to (and close this one)

peterbe:
I approved, pending the Travis tests pass.

peterbe:
There we go! Let's do it!

PR #200, opened 2018-05-13
Remove CLI and make puppeteer peerDependecy

stereobooster:


stereobooster:
> If I use `minimalcss` in `react-snap` it will download two copies of `puppeteer`.

this is how npm works, if you have something as dependency you will get isolated copy of this thing, to reuse packages you need to declare it as peerDependency.

so in `react-snap` and `minimalcss-cli` it is dependency, but in `minimalcss` it is peerDependency, which will make `minimalcss` to use `puppeteer` from `react-snap` or `minimalcss-cli`.

peterbe:
Ok. That I understood. (thought it was more complicated than that)

What I still don't really know is when should you use `peerDependencies` and when should you use `dependencies`. In this PR; **why** use `peerDependencies`? 
Is it to avoid other projects installing `puppeteer` in multiple places in `node_modules`?

stereobooster:
So as of now `minimalcss` and `react-snap` list `puppeteer` dependency as:

```
"puppeteer": "^1.2.0",
```

which get resolved to 

```
puppeteer@^1.2.0:
  version "1.4.0"
```

and installed only once - yay. But if `react-snap` would have `puppeteer` dependency as 

```
"puppeteer": "1.3.0",
```

it would resolve to two different packages

```
puppeteer@1.3.0:
  version "1.3.0"

puppeteer@^1.2.0:
  version "1.2.0"
```

and as a result

```
$ du -sh node_modules/minimalcss/node_modules/* | grep puppeteer
172M	node_modules/minimalcss/node_modules/puppeteer
$ du -sh node_modules/* | grep puppeteer
171M	node_modules/puppeteer
```

---

Where is `peerDependencies` does not install anything, it uses parents project version. So if I would declare in `react-snap`

```
"minimalcss": "^0.7.5",
"puppeteer": "1.3.0",
```

`minimalcss` will not install separate copy of `puppeteer`, but instead will use one from given project e.g. `node_modules/puppeteer` instead of `node_modules/minimalcss/node_modules/puppeteer` (if versions are compatible)

---

Other way to handle this is to keep versions of `puppeteer` in sync in `minimalcss` and `react-snap`. Which is also possible

peterbe:
Another silly question...

If I have a bunch of different programs I'm using in a project. For example, I use `minimalcss` to handle my critical CSS. And I use webpages-to-pdf (I made this up) to handle my PDF maker. If both of these tools depend on puppeteer, does that mean that *I* have to manage that now? What if `webpages-to-pdf` or `my-screenshotter-app` depend and require a certain old version of puppeteer (or a bleeding edge version), will I not be able to have two different `puppeteer`s installed?

stereobooster:
Depends 

If each package declares `dependency` on `puppeteer` and all depend on different version npm/yarn will install 3 versions of `puppeteer`. Pros: versions of `puppeteer` independent. Cons: it installs 3 copies of `puppeteer`

If each package declares `peerDependency` on `puppeteer` you will need to add `puppeteer` dependency and use version which will satisfy all packages. Pros: it installs 1 copy of `puppeteer`. Cons: versions of `puppeteer` limited, you need to figure out what compatible version is

stereobooster:
We also can keep version of `puppeteer` in both packages the same and forget about this `peerDependency` thing. It will also work.

peterbe:
So I think we should opt for the simplest solution of using regular `dependencies`. 
There are so many other juicy improvements in this PR. 

One thing that did occur to me is that now that the cli is gone, it's rather hard to test new features of `minimalcss` on the command line. One idea is to do something like: `cp ~/dev/JAVASCRIPT/minimalcss-cli/cli.js . && node cli.js --version` so that it uses the current `./node_modules` and `src/minimalcss.js`. If that's an acceptable dev/debug technique we can later add a blurb about it in the README.

PR #224, opened 2018-06-20
upgrade filesiz

peterbe:


peterbe:
I don't know why Renovate didn't pick this up.

package.json
```diff
@@ -26,7 +26,7 @@
     "cheerio": "1.0.0-rc.2",
     "css-tree": "1.0.0-alpha.28",
     "csso": "~3.5.0",
-    "filesize": "^3.5.11",
+    "filesize": "^3.6.1",
```
stereobooster: if you will change to exact versions renovate will update it for you

peterbe: Ah! Actually, I'm not sure this was a good idea anyway. By forcing exactly on `==3.6.1` then if someone installs `minimalcss` and other stuff that might use `filesize` too, it will require two installs of that potentially. 

As a matter of fact, there's very little point in forcing an exact version for anything where we don't depend on some particular important feature.

peterbe:
Let's not bother unless there's a particular feature we need that we didn't need before.

Issue #231, opened 2018-07-02
TypeError: Cannot read property 'forEach' of undefined


quasiperfect:
hi

what could trigger that error ?

i got this issue testing on http://codingheads.com/

stereobooster:
Caused by:

```
@-moz-document url-prefix() {
	inputtype:radio],input[type=checkbox]{vertical-align:middle;
	margin-top:-.3em
}
}
```

Which is parsed to

```
{ type: 'Raw',
  loc: null,
  value: 'inputtype:radio],input[type=checkbox]' }
```

quasiperfect:
@stereobooster thanks for your help. that would be great(handle invalid CSS)

harrygreen:
This error actually found a ton of invalid crap in our production CSS!

stereobooster:
fixed in #237

PR #237, opened 2018-07-25
Throw error if invalid CSS is encountered

harrygreen:
This throws an error if invalid CSS is found during `csstree.walk`. 

I couldn't find a nice way to output the line or surrounding nodes to help the dev.

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -456,6 +456,14 @@ const minimalcss = async options => {
           return;
         }
 
+        if (!node.prelude.children) {
+          throw new Error(
+            `Invalid CSS found while evaluating ${href}: "${
+              node.prelude.value
+            }"`
+          );
+        }
+
         node.prelude.children.forEach((node, item, list) => {
           // Translate selector's AST to a string and filter pseudos from it
           // This changes things like `a.button:active` to `a.button`
--- a/tests/main.test.js
+++ b/tests/main.test.js
@@ -145,6 +145,20 @@ test('form elements', async () => {
   expect(finalCss).toMatch('option:selected');
 });
 
+test.only('invalid css', async () => {
+  expect.assertions(1);
+
+  try {
+    await runMinimalcss('invalid-css');
+  } catch (error) {
+    const expectedUrl = 'http://localhost:3000/invalid-css.css';
+    const expectedInvalidCSS = '$body';
+    expect(error.toString()).toMatch(
+      `Invalid CSS found while evaluating ${expectedUrl}: "${expectedInvalidCSS}"`
+    );
+  }
+});
+
 test('handles 307 CSS file', async () => {
   const { finalCss } = await runMinimalcss('307css');
   expect(finalCss).toEqual('p{color:violet}');
--- a/tests/examples/invalid-css.css
+++ b/tests/examples/invalid-css.css
@@ -1,0 +1,3 @@
+$body {
+  color: violet;
+}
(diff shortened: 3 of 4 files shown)


harrygreen:
Hi @stereobooster 

Test added 👍

tests/main.test.js
```diff
@@ -145,6 +145,15 @@ test('form elements', async () => {
   expect(finalCss).toMatch('option:selected');
 });
 
+test('invalid css', async () => {
+  expect.assertions(1);
+  try {
+    await runMinimalcss('invalid-css');
+  } catch (error) {
+    expect(error).toBeDefined();
```
stereobooster: Maybe we can create assertion for the message of the error. WDYT?

src/run.js
```diff
@@ -456,6 +456,10 @@ const minimalcss = async options => {
           return;
         }
 
+        if (!node.prelude.children) {
+          throw new Error(`Invalid CSS found: ${node.prelude.value}`);
```
stereobooster: We can use `href` so it would be easy to understand where the error comes from

tests/main.test.js
harrygreen: Yes, I'll check that instead.

src/run.js
harrygreen: how about:
```js
throw new Error(
  `Invalid CSS found while evaluating ${href}: "${
    node.prelude.value
  }"`
);
```
?

tests/main.test.js
```diff
@@ -145,6 +145,17 @@ test('form elements', async () => {
   expect(finalCss).toMatch('option:selected');
 });
 
+test('invalid css', async () => {
+  expect.assertions(1);
+  try {
+    await runMinimalcss('invalid-css');
+  } catch (error) {
+    expect(
```
stereobooster: why not exact match `expect(error).toEqual('Error: Invalid CSS found: ...');`?

src/run.js
stereobooster: Maybe `SyntaxError at ${href} near "${node.prelude.value}"`, but have no strong opinions

harrygreen:
Hey @stereobooster, I've updated the invalid CSS mock and made the error string test more explicit.

jc275:
Depending on the use case (e.g. accepting unsafe/third-party css) it may be preferable to make this optional. For example, on malformed CSS rule either...
1. throw error
2. remove rule and log warning
3. remove rule silently

Similar errors will still be thrown by `csso`, but there's a PR that allows it to swallow malformed CSS rules. https://github.com/css/csso/pull/381

peterbe:
Do you have a web page I can test this with?

peterbe:
@jc275 ...

> may be preferable to make this optional

I agree but I think it should be a separate PR just for that. Would you like to try adding one? 
However, I would prefer it to be optional to NOT skip broken CSS. Meaning, by default it should throw errors and if you're really certain that that CSS can be ignored, then you override a default setting to skip broken CSS.

peterbe:
@harrygreen Can you change (and rebase) the commit message to say:
`Throw error if invalid CSS is encountered, fixes #231`

harrygreen:
@peterbe > Do you have a web page I can test this with?

peterbe:
Released as part of 0.7.7 https://www.npmjs.com/package/minimalcss

harrygreen:
thanks @peterbe @stereobooster

PR #244, opened 2018-08-01
Handle multiple semicolons to prevent csso crash, fixes #243

### LINKED ISSUE #243: TypeError: Cannot read property '0' of undefined
Repro:
https://minimalcss.app/?url=http://www.nortonrosefulbright.com

Minimal repro:
Add a semicolon in https://css.github.io/csso/csso.html

Caused by `csso.minify`, which fails when the stylesheet contains a double semicolon.
https://github.com/css/csso/issues/378

jc275:
Fixes #243

There may be a safer way to handle this (e.g. via `css-tree`) rather than naively editing the CSS text... but this works for me :)

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -315,7 +315,10 @@ const processPage = ({
             link.media !== 'print' &&
             !link.href.toLowerCase().startsWith('data:')
           ) {
-            hrefs.push(link.href);
+            // Fragments are omitted from puppeteer's response.url(),
+            // so we need to strip them here, otherwise the hrefs
+            // won't always match when we check for missing ASTs.
+            hrefs.push(link.href.split('#')[0]);
           }
         });
         return {
--- a/src/run.js
+++ b/src/run.js
@@ -127,20 +127,20 @@ const processPage = ({
 
     const tracker = createTracker(page);
     const safeReject = error => {
-      if (!fulfilledPromise) {
-        if (error.message.startsWith('Navigation Timeout Exceeded')) {
-          const urls = tracker.urls();
-          if (urls.length > 1) {
-            error.message += `\nTracked URLs that have not finished: ${urls.join(
-              ', '
-            )}`;
-          } else if (urls.length > 0) {
-            error.message += `\nFor ${urls[0]}`;
-          }
+      if (fulfilledPromise) return;
+      fulfilledPromise = true;
+      if (error.message.startsWith('Navigation Timeout Exceeded')) {
+        const urls = tracker.urls();
+        if (urls.length > 1) {
+          error.message += `\nTracked URLs that have not finished: ${urls.join(
+            ', '
+          )}`;
+        } else if (urls.length > 0) {
+          error.message += `\nFor ${urls[0]}`;
         }
-        tracker.dispose();
-        reject(error);
       }
+      tracker.dispose();
+      reject(error);
     };
 
     const debug = options.debug || false;
--- a/src/run.js
+++ b/src/run.js
@@ -218,6 +218,10 @@ const processPage = ({
           redirectResponses[responseUrl] = redirectsTo;
         } else if (resourceType === 'stylesheet') {
           response.text().then(text => {
+            // Double semicolons can crash csso.
+            while (/;\s*;/.test(text)) {
+              text = text.replace(/;\s*;/g, ';');
+            }
             const ast = csstree.parse(text);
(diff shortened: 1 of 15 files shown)


src/run.js
```diff
@@ -218,6 +218,10 @@ const processPage = ({
           redirectResponses[responseUrl] = redirectsTo;
         } else if (resourceType === 'stylesheet') {
           response.text().then(text => {
+            // Double semicolons can crash csso.
```
peterbe: Add a comment with a link to https://github.com/css/csso/issues/378

jc275:
Whoops, guess I should probably have used some fancypants git commands to reduce the commit-spam :)

src/run.js
```diff
@@ -218,6 +218,12 @@ const processPage = ({
           redirectResponses[responseUrl] = redirectsTo;
         } else if (resourceType === 'stylesheet') {
           response.text().then(text => {
+            // Semicolon sequences can crash CSSO, so we remove them.
+            // https://github.com/peterbe/minimalcss/issues/243
+            // https://github.com/css/csso/issues/378
+            while (/;\s*;/.test(text)) {
```
stereobooster: we have file `src/utils.js` you can place this function there with explanations in jsdoc

tests/main.test.js
```diff
@@ -225,6 +225,23 @@ test('accept CSSO options', async () => {
   expect(finalCss).not.toMatch('test css comment');
 });
 
+test('handles extra semicolons', async () => {
+  // Extra semicolons can cause csso.minify() to throw:
+  // [TypeError: Cannot read property '0' of undefined]
+  // https://github.com/peterbe/minimalcss/issues/243
+  // https://github.com/css/csso/issues/378
```
stereobooster: I guess 

```js
const { finalCss } = await runMinimalcss('extra-semicolons');
expect(finalCss).toMatch('a{color:red}');
```

would be enough for test, no need check if exception undefined if we check return value

stereobooster:
Can you please squash commits and change message to:

```
Handle multiple semicolons to prevent csso crash, fixes #243
```

stereobooster:
https://stackoverflow.com/a/5189600 this is what I meant by squash commits. And yes it is also possible to squash at merge via button in the interface

jc275:
I've been editing this directly on github (which no longer seems like a brilliant idea) so I don't have a local branch to rebase/squash. Can I leave the squashing to you?

src/utils.js
```diff
@@ -30,4 +30,18 @@ const unquoteString = string => {
...
+ * github.com/css/csso/issues/378
+ * @param {string} css
+ * @return {string}
+ */
+const removeSequentialSemis = css => {
+  while (/;\s*;/.test(css)) {
+    css = css.replace(/;\s*;/g, ';');
+  }
+  return css;
+};
```
peterbe: Now, if you feel like you want to perfect this PR, since that string function got uplifted into its own function, can you add a simple test for it. 

At the moment the `utils.test.js` doesn't exist but it [soon will hopefully](https://github.com/peterbe/minimalcss/pull/257).

src/run.js
```diff
@@ -218,6 +218,7 @@ const processPage = ({
           redirectResponses[responseUrl] = redirectsTo;
         } else if (resourceType === 'stylesheet') {
           response.text().then(text => {
+            text = utils.removeSequentialSemis(text);
```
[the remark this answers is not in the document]
jc275: https://github.com/peterbe/minimalcss/pull/244/commits/902d2750d956bfc43dc000bebf3516f8ce90aeca

src/utils.js
jc275: I _do_ feel like perfecting this PR, but I have a vague suspicion that creating `utils.test.js` in two separate branches might introduce some kind of merge conflict.

peterbe: If you rebase your branch against origin/master you should have a file now that you can add to.

peterbe: Rebase without squash that is.

jc275: That didn't go as planned...
I think it's going to be simpler for me to submit another pull request, replacing this one.
Squashing into https://github.com/peterbe/minimalcss/compare/master...jc275:remove-sequential-semicolons

jc275:
Sorry for the chaos! Closing in favour of #259. Let's try again...

Issue #240, opened 2018-07-31
[question] What is 'fulfilledPromise' for in 'run.js'?


jc275:
https://github.com/peterbe/minimalcss/blob/915d032efc8bf1c12b87e9cd31f06a971aefeae5/src/run.js#L130

`fulfilledPromise` appears to be *always* `false`. So, I just wondered whether it's supposed to be toggled to `true` under certain conditions (which it currently isn't, as far as I can tell).

peterbe:
Looks like you're right. I can't see where it every sets that to `true`. Can you make a PR and tag me and @stereobooster in it?

jc275:
I'd gladly do that, but I'm not sure what the intended behaviour is! The simplest thing would be to remove the if-condition, since its value never changes. But I guessed that you had something else in mind when creating the `fulfilledPromise` flag.

stereobooster:
```diff
    const safeReject = error => {
      if (fulfilledPromise) return; 
+     fulfilledPromise = true;
      if (error.message.startsWith('Navigation Timeout Exceeded')) {
        const urls = tracker.urls();
        if (urls.length > 1) {
          error.message += `\nTracked URLs that have not finished: ${urls.join(
            ', '
          )}`;
        } else if (urls.length > 0) {
          error.message += `\nFor ${urls[0]}`;
        }
      }
      tracker.dispose();
      reject(error);
    };
```

peterbe:
@jc275 Can you turn @stereobooster 's diff there into a PR?

stereobooster:
resolved in #245

PR #245, opened 2018-08-08
fulfilledPromise

jc275:
Lines 132-143 are moved outside the if-block, but are otherwise unchanged.
https://github.com/peterbe/minimalcss/issues/240


peterbe:
@stereobooster Go ahead and merge.

PR #247, opened 2018-08-08
It seems we gave up idea of TS, so lets remove it

stereobooster:
WDYT? It seems we do not use it. We gave it try, but it seems doesn't work.

Can replace it wit Flow type in the comments. Flow type (at least in comments) works better


peterbe:
Also, something that feels like even bigger fish to fry is to make that break of splitting out minimalcss from minimalcss-cli so that the projects remain smaller and tighter.

PR #248, opened 2018-08-08
Add option timeoutAsWarning

stereobooster:
not sure about option name

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,7 @@ key is `urls`. Other optional options are:
   of strings for headless Chrome](https://peter.sh/experiments/chromium-command-line-switches/).
 * `cssoOptions` - CSSO compress function [options](https://github.com/css/csso#compressast-options)
 * `timeout` - Maximum navigation time in milliseconds, defaults to 30 seconds, pass 0 to disable timeout.
+*  `timeoutAsWarning` - This is workaround for the [bug in puppeteer](https://github.com/GoogleChrome/puppeteer/issues/1908#issuecomment-390214976), when there are no open connections but puppeteer reports it as timeout anyway. If you set it to true minimalcss will report timeouts as warnings instead of rejection. Take a note real timeouts still be reported as errors.
 
 ## Warnings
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -176,6 +176,7 @@ key is `urls`. Other optional options are:
   of strings for headless Chrome](https://peter.sh/experiments/chromium-command-line-switches/).
 * `cssoOptions` - CSSO compress function [options](https://github.com/css/csso#compressast-options)
 * `timeout` - Maximum navigation time in milliseconds, defaults to 30 seconds, pass 0 to disable timeout.
+*  `timeoutAsWarning` - This is workaround for the bug in puppeter, when there are no open connections but puppeteer reports it as timeout anyway. If you set it to true minimalcss will report timeouts as warnings instead of rejection. Take a note real timeouts still be reported as errors.
```
peterbe: "the bug in puppeteer" should be a link to the bug.

src/run.js
```diff
@@ -137,8 +136,14 @@ const processPage = ({
           )}`;
         } else if (urls.length > 0) {
           error.message += `\nFor ${urls[0]}`;
+        } else if (options.timeoutAsWarning) {
+          // This is workaround for the bug in puppeter, when there are
```
peterbe: Again, a URL to the bug here would be super useful.

peterbe: Also, typo on "puppeter".

src/run.js
```diff
@@ -285,7 +285,21 @@ const processPage = ({
       // Second, goto the page and evaluate it with JavaScript.
       // The 'waitUntil' option determines how long we wait for all
       // possible assets to load.
-      response = await page.goto(pageUrl, { waitUntil: 'networkidle0' });
+      try {
+        response = await page.goto(pageUrl, { waitUntil: 'networkidle0' });
```
stereobooster: this will not work, we do not have repsonse

stereobooster:
Ahh it will not work, but it was nice idea

PR #249, opened 2018-08-08
Add ignoreCssError option

stereobooster:
Not sure about name of the option

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,7 @@ key is `urls`. Other optional options are:
   of strings for headless Chrome](https://peter.sh/experiments/chromium-command-line-switches/).
 * `cssoOptions` - CSSO compress function [options](https://github.com/css/csso#compressast-options)
 * `timeout` - Maximum navigation time in milliseconds, defaults to 30 seconds, pass 0 to disable timeout.
+* `ignoreCSSErrors` - By default, any CSS parsing error throws an error in minimalcss. If you know it's safe to ignore (for example, third-party CSS resources), set this to true.
 
 ## Warnings
 
--- a/tests/main.test.js
+++ b/tests/main.test.js
@@ -145,7 +145,7 @@ test('form elements', async () => {
   expect(finalCss).toMatch('option:selected');
 });
 
-test.only('invalid css', async () => {
+test('invalid css', async () => {
   expect.assertions(1);
 
   try {
@@ -157,6 +157,13 @@ test.only('invalid css', async () => {
       `Invalid CSS found while evaluating ${expectedUrl}: "${expectedInvalidCSS}"`
     );
   }
+});
+
+test('ignoreCSSErrors', async () => {
+  const { finalCss } = await runMinimalcss('invalid-css', {
+    ignoreCSSErrors: true
+  });
+  expect(finalCss).toEqual('');
 });
 
 test('handles 307 CSS file', async () => {
--- a/src/run.js
+++ b/src/run.js
@@ -339,7 +339,7 @@ const processPage = ({
 
 /**
  *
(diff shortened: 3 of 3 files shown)


stereobooster:
cc @jc275

tests/main.test.js
```diff
@@ -159,6 +159,13 @@ test.only('invalid css', async () => {
   }
 });
 
+test.only('ignoreCssError', async () => {
+  const { finalCss } = await runMinimalcss('invalid-css', {
+    ignoreCssError: true
+  });
+  expect(finalCss).toEqual('');
```
peterbe: Awesome!

README.md
```diff
@@ -176,6 +176,7 @@ key is `urls`. Other optional options are:
   of strings for headless Chrome](https://peter.sh/experiments/chromium-command-line-switches/).
 * `cssoOptions` - CSSO compress function [options](https://github.com/css/csso#compressast-options)
 * `timeout` - Maximum navigation time in milliseconds, defaults to 30 seconds, pass 0 to disable timeout.
+* `ignoreCssError` - ignore CSS errors. By default minimalcss reports all CSS errors, but if you want to ignore those errors, if you do not have access to those css files, you can set this option to true.
```
peterbe: Sorry for being difficult but that should really be called `ignoreCssErrors` (with a `s` suffix) because the functionality is that it will ignore all CSS errors for the whole duration of the `minimalcss` run.

peterbe: Also, can you uppercase the second instance of the word `css` to `CSS`. For consistency.

peterbe: The ` - ignore CSS errors.` part isn't super useful. The functionality is pretty clear from the name of the option (`ignoreCssErrors`). 

I would change it to something like:

```
By default, any CSS parsing error throws an error in minimalcss. If you know it's safe to ignore (for example, third-party CSS resources), set this to true.
```

stereobooster: Yes sure. I was exactly expecting help on naming

stereobooster:
fixed

README.md
```diff
@@ -176,6 +176,7 @@ key is `urls`. Other optional options are:
   of strings for headless Chrome](https://peter.sh/experiments/chromium-command-line-switches/).
 * `cssoOptions` - CSSO compress function [options](https://github.com/css/csso#compressast-options)
 * `timeout` - Maximum navigation time in milliseconds, defaults to 30 seconds, pass 0 to disable timeout.
+* `ignoreCSSErrore` - By default, any CSS parsing error throws an error in minimalcss. If you know it's safe to ignore (for example, third-party CSS resources), set this to true.
```
peterbe: typo :) "ignoreCSSErrore"

peterbe:
I guess I'll make a new release since this is a new feature.

peterbe:
0.7.8 released.

jc275:
I spotted this far too late, but... what about ignoring JS errors too?

https://github.com/peterbe/minimalcss/blob/9b756476ff03371fdd5c2ace16483f9a1f931b3a/src/run.js#L256

peterbe:
@jc275 Do you have a solution in mind?

jc275:
JS errors could be either logged or thrown, in exactly the same way that CSS errors are. 

CSS-related errors might occur in unexpected places, but at least the on-page JS errors should all be found in one place, as 'pageerror' events. 

I don't know whether there's any advantage to a separate JS error option, so maybe `ignoreCssErrors` could become `ignoreErrors` or `allowErrors`. (After all, if you're trying to extract CSS data from a third-party site, it doesn't really matter what caused the error, since you have no way to avoid it.)

peterbe:
It would be cool to have an `ignoreJSErrors` option for sure. If there's a clear and easy way to implement it so that it doesn't collide with our "own" error handling.

PR #251, opened 2018-08-09
Extracted more utils

stereobooster:


src/run.js
```diff
@@ -164,10 +155,11 @@ const processPage = ({
       }
 
       // A must or else you can't do console.log from within page.evaluate()
-      page.on('console', msg => {
+      page.on('console', async (msg) => {
         if (debug) {
-          for (let i = 0; i < msg.args.length; ++i) {
-            console.log(`${i}: ${msg.args[i]}`);
```
stereobooster: maybe this worked in older versions in puppeteer, but it doesn't work in recent versions. I guess we also need to bump puppeteer version

src/tracker.js
```diff
@@ -23,4 +22,27 @@ const createTracker = page => {
...
+  if (message.startsWith('Navigation Timeout Exceeded')) {
+    const urls = tracker.urls();
+    if (urls.length > 1) {
+      message += `\nTracked URLs that have not finished: ${urls.join(
+        ', '
+      )}`;
+    } else if (urls.length > 0) {
+      message += `\nFor ${urls[0]}`;
+    } else {
+      message += `\nBut there are no pending connections`;
```
stereobooster: added this

src/utils.js
```diff
@@ -30,4 +30,20 @@ const unquoteString = string => {
...
+ * @returns {Promise<Array<any>>}
+ */
+const consoleMessageToArguments = (msg) =>{
+  const text = msg.text();
+  if (text !== 'JSHandle@object') {
+    return Promise.resolve([text]);
+  } else {
+    return Promise.all(msg.args().map(x => x.jsonValue()));
+  }
+});
```
peterbe: Drop the `)` and we'll see if Travis gets happy.

src/utils.js
```diff
@@ -30,4 +30,20 @@ const unquoteString = string => {
...
 };
 
-module.exports = { reduceCSSSelector, unquoteString };
+/**
+ * Convert ConosleMessage to array of values which were passed to console.
+ *
+ * @param {ConosleMessage} msg
+ * @returns {Promise<Array<any>>}
+ */
+const consoleMessageToArguments = (msg) =>{
```
peterbe: Perhaps your `prettier` isn't set up but I bet Travis will complain about this lacking whitespace before the `{`.

peterbe:
I wouldn't worry too much about it. It's a fun side-project and we should focus on features that are fun to have rather than a perfect code base. 

Something that's still on my mind is the organization of a core library and split the cli away from that.

PR #253, opened 2018-08-10
ignoreJSErrors

jc275:
Aside from tests and documentation, this should be the only change needed.

Camel-cased `ignoreJsErrors` (and `ignoreCssErrors`) might be more in keeping with the [style guide](https://github.com/peterbe/minimalcss/blob/master/CONTRIBUTING.md#style-guide).

https://github.com/peterbe/minimalcss/pull/249

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ key is `urls`. Other optional options are:
 * `debug` - all console logging during page rendering are included in the
   stdout. Also, any malformed selector that cause errors in `document.querySelector`
   will be raised as new errors.
-* `skippable` - function wich takes
+* `skippable` - function which takes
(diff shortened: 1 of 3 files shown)


peterbe:
The tests failed but I think it's just bad luck. I restarted them.

jc275:
Awesome, thanks.

stereobooster:
As you know my main use case is SPA (mostly React), so for my case this feature doesn't make sense. If there is JS error it means the user experience broken, but I agree it can be useful for some users of minimalcss.

peterbe:
What's missing is that this new option needs to be explained and mentioned in the README.

peterbe:
Pushed out version 0.7.9 with this new feature.

PR #257, opened 2018-08-16
tests for utility functions

peterbe:
I didn't like that we had extracted out some utility functions but never had *any* tests for them. 
The most important tests are the tests that use the public API but if the code was sufficiently complex that utility functions had to be broken out, we should have at least some tests for them.


peterbe:
Thanks

PR #259, opened 2018-08-17
Handle multiple semicolons to prevent csso crash, fixes #243

### LINKED ISSUE #243: TypeError: Cannot read property '0' of undefined

jc275:
As discussed in https://github.com/peterbe/minimalcss/pull/244. All commits from that branch are merged and squashed into this one, along with a test in `utils.test.js`.


peterbe:
I released version 0.7.10 to celebrate this. Thanks again for all your hard work!

lahmatiy:
FYI This issue was fixed in CSSO 4.0.2 – so a workaround is not needed anymore.

Issue #114, opened 2018-01-26
Probably parse all inline style blocks too


peterbe:
([Originally posted here](https://github.com/peterbe/minimalcss/pull/58#issuecomment-356730109))

----------

I actually thing we should keep this open. The use case isn't a top priority but it's still something we have a great opportunity to solve. Perhaps:
```
$ ./bin/minimalcss.js --inlines URL
```
And it could *just* download each `style` block, using `cheerio`, and compare it against the DOM(s). We're really close to being able to do this.

Also, since we have access to the `document` in puppeteer we have access to `document.styleSheets`.

The question is, what should the "API" (and cli parameters) be for this?

* We could **combine** the inline blocks with all the downloaded external stylesheets into one blob of CSS.
* We could make it ONLY analyze the inline blocks. 
* (Primarily only a concern for the cli) We could make it pretty-print each inline block with a comment like `/* Inline block number 3 */`

Perhaps we start with a basic implementation and worry about the CLI later. E.g.
```javascript
minimalcss
  .minimize({ urls: ['http://peterbecom.dev/css-blocking/ultra-basic.html'], inlines: true })
  .then(result => {
    result.inlines.forEach((block, i) => {
       console.log(`/* Inline style block number ${i + 1} */`)
       console.log(block.finalCss)
       console.log(`/* Used to be ${block.stylesheetContent.length}bytes but is now ${block.finalCss.length} */`)
    })
  })
```

If a site has (and this is not unusual) 1 external `somefatframework.css` and an inline style block, it's extremely unlikely that the bytes savings on the inline style block will be impressive (or even worth it!) but it's nice to see if you have styles in an inline style block that is now never ever used and can be cleaned up. 

----------

Another reason why this might be important is that you might have this:
```html
<html>
<link rel="stylesheet" href="foo.css">
<style>
div.something {
    animation: rotate-dawg infinite 5s;
    font-family: 'Gill Sans', sans-serif;
}
</style>
```
and in `foo.css` you have this:

```css
@keyframe rotate-dawg {
  from {...}
  to {...}
}
@font-face {
    font-family: Gill Sans;
    src: url(....);
}
/* No mention of div.something in this file */
```
Today, if you analyzed that it'd find that the keyframe and the font-face can be removed before there are no CSS selectors, in any of the **external stylesheets**, that needs it.

peterbe:
I wonder, if we can use devtools coverage reporting instead, it'll analyze all inline styles too, no?

[...]

Issue #13, opened 2017-09-21
Use `forceMediaMerge` in csso


peterbe:
csso [3.2.0](https://github.com/css/csso/releases/tag/v3.2.0) introduced `forceMediaMerge` as an aggressive option that tries to merge media rules.

PR #260, opened 2018-08-18
parse style tags

jc275:
- [x]  Adds `options.styletags` to parse any on-page `<style>` tags (in addition to external stylesheets).
- [x]  Processes style tag content exactly the same way as external stylesheet text.
- [x]  Preserves relative ordering of external stylesheets and on-page style tags.
- [x]  Includes `<style>` tags that were created/modified after page load.
- [x]  Does not consider any inline `style="..."` attributes.
- [ ]  Probably fixes https://github.com/peterbe/minimalcss/issues/114

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -107,6 +107,49 @@ const postProcessOptimize = ast => {
   });
 };
 
+const processStylesheet = ({
+  text,
+  pageUrl,
+  responseUrl,
+  stylesheetAsts,
+  stylesheetContents
+}) => {
+  // Semicolon sequences can crash CSSO,
+  // so we remove them from the CSS text.
+  // https://github.com/peterbe/minimalcss/issues/243
+  // https://github.com/css/csso/issues/378
+  text = utils.removeSequentialSemis(text);
+  const ast = csstree.parse(text);
+  csstree.walk(ast, node => {
+    if (node.type !== 'Url') return;
+    const value = node.value;
+    let path = value.value;
+    if (value.type !== 'Raw') {
+      path = path.substr(1, path.length - 2);
+    }
+    const sameHost = url.parse(responseUrl).host === url.parse(pageUrl).host;
+    if (/^https?:\/\/|^\/\/|^data:/i.test(path)) {
+      // do nothing
+    } else if (/^\//.test(path) && sameHost) {
+      // do nothing
+    } else {
+      const resolved = new url.URL(path, responseUrl);
+      if (sameHost) {
+        path = resolved.pathname + resolved.search;
+      } else {
+        path = resolved.href;
+      }
+      if (value.type !== 'Raw') {
+        value.value = `"${path}"`;
+      } else {
+        value.value = path;
+      }
+    }
+  });
+  stylesheetAsts[responseUrl] = ast;
+  stylesheetContents[responseUrl] = text;
+};
+
 const processPage = ({
   page,
   options,
@@ -218,42 +261,13 @@ const processPage = ({
           redirectResponses[responseUrl] = redirectsTo;
         } else if (resourceType === 'stylesheet') {
           response.text().then(text => {
-            // Semicolon sequences can crash CSSO,
-            // so we remove the from the CSS text.
-            // https://github.com/peterbe/minimalcss/issues/243
-            // https://github.com/css/csso/issues/378
-            text = utils.removeSequentialSemis(text);
-            const ast = csstree.parse(text);
-            csstree.walk(ast, node => {
-              if (node.type === 'Url') {
-                let value = node.value;
-                let path = value.value;
-                if (value.type !== 'Raw') {
-                  path = path.substr(1, path.length - 2);
-                }
-                const sameHost =
-                  url.parse(responseUrl).host === url.parse(pageUrl).host;
-                if (/^https?:\/\/|^\/\/|^data:/i.test(path)) {
-                  // do nothing
-                } else if (/^\//.test(path) && sameHost) {
-                  // do nothing
-                } else {
-                  const resolved = new url.URL(path, responseUrl);
-                  if (sameHost) {
-                    path = resolved.pathname + resolved.search;
-                  } else {
-                    path = resolved.href;
-                  }
-                  if (value.type !== 'Raw') {
-                    value.value = `"${path}"`;
-                  } else {
-                    value.value = path;
-                  }
-                }
-              }
+            processStylesheet({
+              text,
+              pageUrl,
+              responseUrl,
+              stylesheetAsts,
+              stylesheetContents
             });
-            stylesheetAsts[responseUrl] = ast;
-            stylesheetContents[responseUrl] = text;
           });
         }
       });
--- a/src/run.js
+++ b/src/run.js
@@ -315,34 +315,76 @@ const processPage = ({
         );
       }
       const evalWithJavascript = await page.evaluate(() => {
+        const html = document.documentElement.outerHTML;
         // The reason for NOT using a Set here is that that might not be
         // supported in ES5.
         const hrefs = [];
-        // Loop over all the 'link' elements in the document and
-        // for each, collect the URL of all the ones we're going to assess.
-        Array.from(document.querySelectorAll('link')).forEach(link => {
-          if (
-            link.href &&
-            (link.rel === 'stylesheet' ||
-              link.href.toLowerCase().endsWith('.css')) &&
-            !link.href.toLowerCase().startsWith('blob:') &&
-            link.media !== 'print' &&
-            !link.href.toLowerCase().startsWith('data:')
-          ) {
-            // Fragments are omitted from puppeteer's response.url(),
-            // so we need to strip them here, otherwise the hrefs
-            // won't always match when we check for missing ASTs.
-            hrefs.push(link.href.split('#')[0]);
+        const styles = [];
+        const isCssStyleTag = elem =>
+          elem.tagName === 'STYLE' &&
+          (!elem.type || elem.type.toLowerCase() === 'text/css');
+        const isStylesheetLink = elem =>
+          elem.tagName === 'LINK' &&
+          elem.href &&
+          (elem.rel.toLowerCase() === 'stylesheet' ||
+            elem.href.toLowerCase().endsWith('.css')) &&
+          !(
+            elem.href.toLowerCase().startsWith('data:') ||
+            elem.href.toLowerCase().startsWith('blob:') ||
+            elem.media.toLowerCase() === 'print'
+          );
+        // #fragments are omitted from puppeteer's response.url(), so
+        // we need to strip them from stylesheet links, otherwise the
+        // hrefs won't always match when we check for missing ASTs.
+        const defragment = href => href.split('#')[0];
+        const pageUrl = defragment(window.location.href);
+        // Create a unique identifier for each style tag by appending
+        // an xpath-like fragment to the page URL.  This allows us to
+        // preserve the relative ordering of external stylesheets and
+        // inline style tags.
+        const styleTagUri = () => `${pageUrl}#style[${styles.length}]`;
+        // Loop over all 'link' and 'style' elements in the document,
+        // in order of appearance. For each element, collect the URI
+        // of all the ones we're going to assess. For style elements,
+        // also extract each tag's content.
+        Array.from(document.querySelectorAll('link, style')).forEach(elem => {
+          if (isStylesheetLink(elem)) {
+            const href = defragment(elem.href);
+            hrefs.push(href);
+          } else if (isCssStyleTag(elem)) {
+            const href = styleTagUri();
+            const text = elem.innerHTML;
+            styles.push({ href, text });
+            hrefs.push(href);
           }
         });
-        return {
-          html: document.documentElement.outerHTML,
-          hrefs
-        };
+        return { html, hrefs, styles };
       });
 
       const htmlWithJavascript = evalWithJavascript.html;
       doms.push(cheerio.load(htmlWithJavascript));
+
+      if (options.styletags) {
+        // Parse each style tag as if it were an external stylesheet.
+        evalWithJavascript.styles.forEach(({ href, text }) => {
+          processStylesheet({
+            text,
+            pageUrl,
+            stylesheetAsts,
+            stylesheetContents,
+            responseUrl: href
+          });
+        });
+      } else {
+        // Remove each style tag URI from the list of hrefs.
+        evalWithJavascript.styles.forEach(({ href }) => {
+          evalWithJavascript.hrefs.splice(
+            evalWithJavascript.hrefs.indexOf(href),
+            1
+          );
+        });
+      }
+
       evalWithJavascript.hrefs.forEach(href => {
         // The order of allHrefs is important! That's what browsers do.
         // But we can't blindly using allHrefs.push() because the href
@@ -365,7 +407,7 @@ const processPage = ({
 
 /**
  *
- * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any, puppeteerArgs: Array<string>, cssoOptions: Object, ignoreCSSErrors?: boolean, ignoreJSErrors?: boolean }} options
+ * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any, puppeteerArgs: Array<string>, cssoOptions: Object, ignoreCSSErrors?: boolean, ignoreJSErrors?: boolean, styletags?: boolean }} options
  * @return Promise<{ finalCss: string, stylesheetContents: { [key: string]: string } }>
  */
 const minimalcss = async options => {
--- a/src/run.js
+++ b/src/run.js
@@ -188,6 +188,7 @@ const processPage = ({
 
     const debug = options.debug || false;
     const loadimages = options.loadimages || false;
+    const styletags = options.styletags || false;
     const withoutjavascript =
       options.withoutjavascript === undefined
         ? true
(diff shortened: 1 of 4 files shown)


jc275:
I know this is probably a massive oversimplification of what you guys were planning, but I thought it might be a good start for (what I assume is) the most likely use case.

tests/main.test.js
```diff
@@ -73,6 +73,13 @@ test('cares only about external CSS files', async () => {
   expect(finalCss).toEqual('.external{color:red}');
 });
 
+test('cares about style tags and external CSS files', async () => {
+  const { finalCss } = await runMinimalcss('css-in-js', {
+    styletags: true
+  });
+  expect(finalCss).toEqual('.cssinjs1,.external,.inline{color:red}');
```
stereobooster: idea is to take into account inline styles so we would know if we can safely remove `@font` and `@keyframes`, but not output as part of css, because those styles already part of html.

stereobooster:
Hi @jc275, thanks for contribution. You can ask if scope of some tickets is not clear. I left one comment (above) which summarises idea of this feature (at least from my point of view). If you have more questions don't hesitate to ask.

jc275:
Thanks for your comments, @stereobooster.

First, as this is an _unsolicited_ PR, I won't be offended if you think it's unconstructive and decide to close it!

Otherwise, a few clarifications if I may...

1. By 'inline styles'... are we talking about `<style>` tags, or `style="..."` attributes, or both?

2. Since the html will always have to be modified—at least, to replace the original stylesheet links—is it essential to keep inline styles where they are? I imagine that a common use case would be to merge `largely-unused-huge-boilerplate-framework.min.css` with a few custom stylesheets or style tags—which may or may not be critical-rendering-path CSS that needs to stay on the page. On that basis, would it be helpful to use an approach like [isomorphic-style-loader](https://github.com/kriasoft/isomorphic-style-loader) or [penthouse](https://github.com/pocketjoso/penthouse) (which [addyosmani/critical](https://github.com/addyosmani/critical) uses under the hood) to decide—after `minimalcss`-ification—which styles should be inlined?

(My main use case is extracting style rules from third-party websites, so I would prefer to combine inline styles with external ones. I realise that the intended purpose is for users to optimise their own websites/apps, but it seems to me that merging linked stylesheets with inline styles may sometimes be very useful for that use case, too.)

stereobooster:
> By 'inline styles'... are we talking about <style> tags, or style="..." attributes, or both?

Both because: `<div style="animation-name: example">` and `<h1 style="font-family: GoogleFont;">`. To detect if we need `@font` and `@keyframes` we need to parse both (without flag). But if we want also inline styles to be part of the output, I would say this should be behind the flag (what you did), but I'm still not sure about use case.

jc275:
I can't really offer any help with `@keyframes` and `@font` rules. For my specific use case, I'm only interested in other types of CSS rules.

I _can_ see a clear, common use case for merging (and perhaps splitting, based on critical render path) the linked and inline styles. However, I haven't considered `style` attributes at all, so my PR doesn't fully solve that problem.

As an aside, would it be cleaner if everything currently happening inside puppeteer's `page.evaluate(() => {...})` were simplified to:
```js
const htmlWithJavascript = await page.evaluate(() => document.documentElement.outerHTML);
const $ = cheerio.load(htmlWithJavascript);
doms.push($);
// do all the interesting stuff in the local scope, instead of the browser
$('link, style, [style]').each(utils.someHelperFunction);
```

jc275:
Maybe it's necessary to repeat the whole process multiple times, independently minimizing the css for external stylesheets, inline styles, and both combined. Then you can do (pseudocode):
```js
// once all unused css rules have been removed...
outputRules = externalAndInlineRules.filter(rule =>
  externalRules.includes(rule) // current behaviour
  || !inlineRules.includes(rule) // for external @font and @keyframes
);
```
(I'm just thinking aloud, as it's too much for me to take on.)

peterbe:
I haven't had a chance yet to look at the code but I skimmed it really quick. This is so exciting!

peterbe:
You've done an excellent job but I think we need to break it down a little bit. 

@stereobooster is right. The primary objective is to not remove things like `@font` based *only* on all the external style sheets. For that, you shouldn't need to explicitly use flags/options like `--styletags`. It should just work automatically. 

I would advise that we break up your awesome code into two PRs. 

## 1. One PR just to survive these two unit tests:

```html
<html>
<link rel=stylesheet href="hasfonts.css">
<style>p { font-family: Lato; }</style>
</html>
```
...where `hasfonts.css` contains...

```css
@font-face {
  font-family: 'Lato';
  font-style: normal;
  font-weight: 400;
  src: local('Lato Regular'), local('Lato-Regular'), url(https://fonts.gstatic.com/s/lato/v14/S6uyw4BMUTPHjxAwXiWtFCfQ7A.woff2) format('woff2');
  unicode-range: U+0100-024F;
}
```

Another example of a unit test would be this:
```html
<html>
<link rel=stylesheet href="hasanimations.css">
<body>
<p style="animation:progress-bar-stripes 2s linear infinite">Hej</p>
</body>
</html>
```

...where `hasanimations.css` contains this:
```css
@keyframes progress-bar-stripes {
  from {
    background-position: 40px 0;
  }
  to {
    background-position: 0 0;
  }
}
```

To solve this, we'd need to pass two AST objects to [`postProcessOptimize`](https://github.com/peterbe/minimalcss/blob/7a43191b27a199da736261eaba56aa1a52472b66/src/run.js#L33). One AST (that gets modified) and one that we inspect. 
So, for example, instead of:
```javascript
  const activeAnimationNames = new Set(
    csstree.lexer
      .findAllFragments(ast, 'Type', 'keyframes-name')
      .map(entry => csstree.generate(entry.nodes.first()))
  );

  csstree.walk(ast, {
    visit: 'Atrule',
    enter: (node, item, list) => {
      const basename = csstree.keyword(node.name).basename;
      if (basename === 'keyframes') {
        if (!activeAnimationNames.has(csstree.generate(node.prelude))) {
          list.remove(item);
        }
      } else if (basename === 'media') {
        if (csstree.generate(node.prelude) === 'print') {
          list.remove(item);
        }
      }
    }
  });
```
perhaps something like this:
```javascript
  const activeAnimationNames = new Set(
    csstree.lexer
      .findAllFragments(astComplete, 'Type', 'keyframes-name')  // HERE! Note the use of 'astComplete'
      .map(entry => csstree.generate(entry.nodes.first()))
  );

  csstree.walk(ast, {
    visit: 'Atrule',
    enter: (node, item, list) => {
      const basename = csstree.keyword(node.name).basename;
      if (basename === 'keyframes') {
        if (!activeAnimationNames.has(csstree.generate(node.prelude))) {
          list.remove(item);
        }
      } else if (basename === 'media') {
        if (csstree.generate(node.prelude) === 'print') {
          list.remove(item);
        }
      }
    }
  });
```
Note here that in the first block we generate the set `activeAnimationNames` based on `astComplete` and in the second block we modify (in place) an object called `ast`.

To do that, we need to create two AST objects and later call:
```javascript
postProcessOptimize(allCombinedAst, allCompleteCombinedAst);
```
...[here](https://github.com/peterbe/minimalcss/blob/7a43191b27a199da736261eaba56aa1a52472b66/src/run.js#L551).

That "complete AST" (called `allCompleteCombinedAst` in the example above), needs to be created not only by downloading all external style sheets but also every single `<style>` tag **and** every `style` attribute. 

The tricky thing is that if you add the `<style>` tag's CSS payloads to the AST, that's valid CSS. But in the case of `<p style="animation:progress-bar-stripes 2s linear infinite">` the value is just `animation:progress-bar-stripes 2s linear infinite` which isn't valid CSS. So you'd have to pretend that it's actually `p{  animation:progress-bar-stripes 2s linear infinite  }`.

## 2. One PR just to extract all the `<style>` tags

I think this would be easier. But perhaps it needs to depend on code from the 1. point above. It could sit under an option/flag like `--inlinestyles` It would basically, make that AST based on something like `$('link, style').each(utils.someHelperFunction);` instead of `$('link').each(utils.someHelperFunction);` as per your example. 

I could see myself using this. That would make it possible to replace ALL `<link>` tags' content **and** all `<style>` tags' content into one single `<style>` tag.

peterbe:
> Maybe it's necessary to repeat the whole process multiple times, independently minimizing the css for external stylesheets, inline styles, and both combined.

I think this is wrong. The point, made above, is that you might have actual connections between the CSS in the external stylesheets (defining things like keyframes) and the inline stylesheets (referring to keyframe animations by name).

peterbe:
> As an aside, would it be cleaner if everything currently happening inside puppeteer's `page.evaluate(() => {...})` were simplified to: ...

You're referring to [this code](https://github.com/peterbe/minimalcss/blob/7a43191b27a199da736261eaba56aa1a52472b66/src/run.js#L304-L340), right? 
That's an excellent point! I suspect the code is the way it is because in the early days we didn't have `cheerio`. The job of that code is to get the whole HTML *and* the list of external stylesheet `hrefs`. That could easily be done outside of puppeteer in a utility function that uses cheerio. 

That would make for an excellent PR on the side as a nice refactoring step.

jc275:
Wow, thanks for the comments and analysis.

Regarding `cheerio`, refactoring from `page.evaluate(...)` might produce more modular, testable code. But, on the other hand, `cheerio` has to iterate over non-standard POJO representations of each node, while chromium has native access to each `HTMLElement`. There are some advantages to the native DOM API (such as missing attributes defaulting to empty strings rather than `undefined`) which might allow us to keep the code simpler and more readable. So I'm in two minds about this.


> ## 2. One PR just to extract all the `<style>` tags
If I've understood you correctly, that's exactly what the current PR is. (I called the option `styletags` instead of `inlinestyles` because it doesn't capture any inline `style` attributes.)

jc275:
> I would advise that we break up your awesome code into two PRs.
> ## 1. One PR just to survive these two unit tests
My awesome code doesn't address this at all :)

jc275:
> So you'd have to pretend that it's actually `p { animation:progress-bar-stripes 2s linear infinite }`.

That could lead to problems if one `p` supersedes another.
Maybe create dummy IDs like this?
```css
#__minimalcss_inline_1__ { animation: progress-bar-stripes 2s linear infinite }
#__minimalcss_inline_2__ { animation: progress-bar-stripes 1s ease }
```

peterbe:
> If I've understood you correctly, that's exactly what the current PR is. 

I unchecked that this PR fixes #114 in the PR description. Right?

I'll review it properly when time allows. 

We'll have to make another attempt at solving the problem of checking if a `@font-face` is mentioned in the style tags in a separate PR after this.

jc275:
https://github.com/peterbe/minimalcss/blob/7a43191b27a199da736261eaba56aa1a52472b66/src/run.js#L304-L305
`Set` should work in chromium; I think an array is still needed because we need to preserve the ordering of stylesheets (and style tags).

peterbe:
I've finally had time to look into this. 
First of all, you need to add this new option to `bin/minimalcss.js` so it becomes possible to do something like `--include-styletags` or something.

Secondly, I'm not sure I understand how it works. I first ran:
```
node bin/minimalcss.js -o /tmp/peterbecom.master.css https://www.peterbe.com/
```
using the `master` branch. 

Then, I checked out your branch, and before using it I manually edited to say:
```
options.styletags=true;
```
...so I can keep testing with the cli. 
Now I run:
```
node bin/minimalcss.js -o /tmp/peterbecom.css https://www.peterbe.com/
```

The one made with your branch is bigger:
```
▶ ls -l /tmp/peterbecom.*css
-rw-r--r--  1 peterbe  wheel  24876 Sep  5 20:38 /tmp/peterbecom.css
-rw-r--r--  1 peterbe  wheel  22910 Sep  5 20:38 /tmp/peterbecom.master.css
```
The difference is 1966 bytes. I *would* expect it to be 108 bytes, because that site is weird. 
It uses a bunch of `<link rel="stylesheet">` tags but it also has one `<style>` block which is never in an external style sheet. That one contains:
```
h2{margin-top:60px;padding-top:60px;border-top:1px solid #e8e9e9;font-size:2rem}div.post{margin-bottom:60px}
```
which is 108 bytes long. 

I took the two `.css` files and uploaded them here:
(using master branch) https://gist.github.com/peterbe/80332a86f9153667c4f07f6c8c0dbfb1
and
(using your branch) https://gist.github.com/peterbe/ace6c2e739b44d58675e975237076208
Before uploading them, I ran them both through `prettier` to make sense of it. 
If you look at the CSS you branch produced `div.post {...}`, for example, which was only in a style tag. 

But what's up with all the other differences?

peterbe:
Note, https://www.peterbe.com was perhaps a weird example to use but it just happened to be the first one I tried. That site actually uses `minimalcss` to extract all the external stylesheets into a block of minimal CSS and then inserts that into the HTML with a `<style>` tag and has all the remaining external CSS loaded async and after the DOM is loaded.

peterbe:
It worked better with `https://www.buzzfeednews.com/article/daveyalba/facebook-philippines-dutertes-drug-war`. Just a random URL. 
That page has a big fat `<style>` block which is 107,188 bytes long but after processing with your branch the whole thing is reduced down to 34,538 bytes.

jc275:
Regarding the differences you see on peterbe.com, isn't that because of the second (much larger) style tag?

peterbe:
> Regarding the differences you see on peterbe.com, isn't that because of the second (much larger) style tag?

Here's the thing. That site, when rendered in the browser will kinda have two copies of a bunch of things. The `<style>` block has all the critical CSS extracted from files like `https://www.peterbe.com/static/css/base.min.8c1ef15190f7.css`. Once that file has been loaded it should contain everything the `<style>` block in the HTML *contained*. 
E.g.

```html
<head>
<link rel="stylesheet" href="main.css" lazy-load=true>
<style>
A { ... }
C { ... }
E { ... }
</style>
</head>
```
and `main.css` contains:
```css
A { ... }
B { ... }
C { ... }
D { ... }
E { ... }
F { ... }
```

So, once merged it should only become:
```css
A { ... }
B { ... }
C { ... }
D { ... }
E { ... }
F { ... }
```
Nothing more. Curious.

jc275:
After a bit of re-ordering, the diff between your two gists ends up looking like this:
```css
/*! License for minified and inlined CSS originally belongs to Semantic UI. See individual files in https://github.com/peterbe/django-peterbecom/tree/master/peterbecom/base/static/css */
@media only screen and (max-width: 767px) {
  .ui.container {
    width: auto !important;
    margin-left: 1em !important;
    margin-right: 1em !important;
  }
}
@media only screen and (min-width: 768px) and (max-width: 991px) {
  .ui.container {
    width: 723px;
    margin-left: auto !important;
    margin-right: auto !important;
  }
}
@media only screen and (min-width: 992px) and (max-width: 1199px) {
  .ui.container {
    width: 933px;
    margin-left: auto !important;
    margin-right: auto !important;
  }
}
@media only screen and (min-width: 1200px) {
  .ui.container {
    width: 1127px;
    margin-left: auto !important;
    margin-right: auto !important;
  }
}
@media only screen and (max-width: 767px) {
  .ui.segment:not(.vertical) .ui.stackable.page.grid {
    margin-left: -1rem !important;
    margin-right: -1rem !important;
  }
}
@media only screen and (max-width: 767px) {
  .ui.menu > .ui.container {
    width: 100% !important;
    margin-left: 0 !important;
    margin-right: 0 !important;
  }
}
@media only screen and (min-width: 768px) {
  .ui.menu:not(.secondary):not(.text):not(.tabular):not(.borderless)
    > .container
    > .item:not(.right):not(.borderless):first-child {
    border-left: 1px solid rgba(34, 36, 38, 0.1);
  }
}
@media (max-width: 414px) {
  h1.ui.header {
    font-size: 2.5rem;
  }
}
@media (min-width: 700px) {
  pre {
    padding: 10px;
  }
}
@media (min-width: 767px) {
  .main.menu .right.dropdown {
    display: none !important;
  }
  .main.menu .right.menu {
    display: flex;
  }
}
@media (max-width: 768px) {
  .main.menu .right.dropdown {
    display: flex !important;
  }
  .main.menu .right.menu {
    display: none !important;
  }
}
h2 {
  margin-top: 60px;
  border-top: 1px solid #e8e9e9;
}
div.post {
  margin-bottom: 60px;
}
```
Not quite sure what to make of it. Any thoughts on where all the `@media` queries are coming from?

As an aside, it doesn't look as if they're minified very effectively, as there's a lot of duplication, e.g.
```css
@media only screen and (max-width: 767px) {
  .ui.container { /* ... */ }
}
@media only screen and (max-width: 767px) {
  .ui.segment:not(.vertical) .ui.stackable.page.grid { /* ... */ }
}
@media only screen and (max-width: 767px) {
  .ui.menu > .ui.container { /* ... */ }
}
```

peterbe:
A ha! I think I know what means. [Merging media queries is dangerous](https://github.com/css/csso/issues/162)
That's why the [`forceMediaMerge`](https://github.com/css/csso#compressast-options) is disabled by default. 

In https://github.com/peterbe/minimalcss/issues/13 I wanted it to be an option for the brave souls that dare to pass `forceMediaMerge` into `csso.compress()`. 

But it's probably best to leave that as is for now. I certainly wouldn't dare to enable it for my site. 

In conclusion. Your branch is good and healthy!! Ignore testing with www.peterbe.com.

So my only outstanding issue is that your new option needs to extended in `bin/minimalcss.js` so one can use this cool new feature on the command line.

src/run.js
```diff
@@ -301,34 +315,76 @@ const processPage = ({
...
-          html: document.documentElement.outerHTML,
-          hrefs
-        };
+        return { html, hrefs, styles };
       });
 
       const htmlWithJavascript = evalWithJavascript.html;
       doms.push(cheerio.load(htmlWithJavascript));
+
+      if (options.styletags) {
```
peterbe: Can you move this up to where all the other `options` are dealt with. E.g.

```javascript
const styletags = options.styletags || false;

....

if (styletags) {
...
```

peterbe: Mainly for consistency and that it feels like we're done with `options` after it has been turned into variables.

tests/main.test.js
peterbe: Can you add a comment here to explain the awesomeness. 
It could be something like:
```
// The css-in-js.html fixture has an external stylesheet and a <style> 
// tag. The <style> tag contains the selector `.inline{...}`.
// This test asserts that the selectors from the inline <style> tags really
// are included in the final CSS. 

```

src/run.js
jc275: Done.

PR #265, opened 2018-09-07
Disable Service Workers

### LINKED ISSUE #223: Disable Service Workers
Service Workers are a thing to help the user agent come back. Then, in the hope of not having to download a bunch of stuff over the network again. Just like caching. Since minimalcss starts a new puppeteer Browser "each time" (`minimalcss-server` uses some Browser recycling), there is never any benefit of using a service worker. Disabling it might help reduce bugs and a small performance boost since it doesn't need to consider all the assets that the service worker dictates.

peterbe:
Fixes #223 

@stereobooster Do you know of a good way to test this? Perhaps it's not possible. Perhaps it's not particularly important. 

Also, [here's how it used to fail](https://github.com/peterbe/minimalcss/issues/223#issuecomment-419529682)

Now, when you run the same you get:

```bash
▶ node bin/minimalcss.js https://googlechrome.github.io/samples/service-worker/basic/
body{background-color:#fff;box-sizing:border-box;font-family:"Roboto","Helvetica","Arial",sans-serif}@media screen and (min-width:832px){body{width:800px;margin:0 auto}}h1{margin-bottom:-.3em}h3{margin-bottom:-.2em;margin-top:2em}.availability{margin-bottom:2em}.highlight{border-radius:.75em;border:1px solid #f0f0f0;display:block;margin:.5em;overflow-x:auto;padding:.5em}code{font-family:Inconsolata,Consolas,monospace}.k,.o{font-weight:700}.c1,.cm{color:#998;font-style:italic}.kd{font-weight:700}.nb{color:#0086b3}.mi{color:#099}.s1{color:#d14}#container{display:flex;flex-direction:row;flex-wrap:wrap;justify-content:space-around}li{margin-bottom:1em}
```

And if you're nuts and you like errors you can get them back:

```bash
▶ node bin/minimalcss.js --enableserviceworkers https://googlechrome.github.io/samples/service-worker/basic/
[Error: DOMException: Failed to register a ServiceWorker: No URL is associated with the caller's document.]
```

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -408,14 +408,18 @@ const processPage = ({
 
 /**
  *
- * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any, puppeteerArgs: Array<string>, cssoOptions: Object, ignoreCSSErrors?: boolean, ignoreJSErrors?: boolean, styletags?: boolean }} options
+ * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any, puppeteerArgs: Array<string>, cssoOptions: Object, ignoreCSSErrors?: boolean, ignoreJSErrors?: boolean, styletags?: boolean, enableServiceWorkers?: boolean }} options
  * @return Promise<{ finalCss: string, stylesheetContents: { [key: string]: string } }>
  */
 const minimalcss = async options => {
   const { urls } = options;
   const debug = options.debug || false;
   const cssoOptions = options.cssoOptions || {};
+  const enableServiceWorkers = options.enableServiceWorkers || false;
   const puppeteerArgs = options.puppeteerArgs || [];
+  if (!enableServiceWorkers) {
+    puppeteerArgs.push('--enable-features=NetworkService');
+  }
   const browser =
     options.browser ||
     (await puppeteer.launch({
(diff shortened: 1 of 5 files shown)


peterbe:
Also, this upgrade of requiring at least puppeteer 1.8.0 might be too big a change. 1.7 should be enough to have that `--enable-features=NetworkService` option. 

How do you feel about this bold upgrade?

peterbe:
Ping @stereobooster if you have time. It's probably not perfect but the example is clear that it does solve a broken problem.

peterbe:
Before:
screen shot 2018-09-13 at 9 50 15 am

After:
screen shot 2018-09-13 at 10 47 03 am

peterbe:
Note, because `minimalcss-server` makes its own browser instance, I had to [change it too](https://github.com/peterbe/minimalcss-server/commit/d4661ef4e6aaffa5f2ee1986ff07a1bb91404098#diff-78c12f5adc1848d13b1c6f07055d996eR28).

src/run.js
```diff
@@ -408,14 +408,18 @@ const processPage = ({
...
  * @return Promise<{ finalCss: string, stylesheetContents: { [key: string]: string } }>
  */
 const minimalcss = async options => {
   const { urls } = options;
   const debug = options.debug || false;
   const cssoOptions = options.cssoOptions || {};
+  const enableServiceWorkers = options.enableServiceWorkers || false;
   const puppeteerArgs = options.puppeteerArgs || [];
+  if (!enableServiceWorkers) {
+    puppeteerArgs.push('--enable-features=NetworkService');
```
stereobooster: FYI there are some issues with this flag, it is not fully ready https://github.com/GoogleChrome/puppeteer/issues/3471

As soon as it will be ready it will be enabled by default

Issue #274, opened 2018-10-17
Found stylesheets that failed to download


nicolas-t:
Hello,

I get the error `Found stylesheets that failed to download`

You can reproduce the error with this command :
`./node_modules/.bin/minimalcss https://bazaar.co/hello/ > minimal.min.css`

You'll see that the css files in question are accessible in your browser, yet minimalcss can't download them.

Should I change the configuration of my server ? 

I've first opened an issue here : https://github.com/stereobooster/react-snap/issues/275 
But I think it's more related to minimalcss.

Thanks

nicolas-t:
> Why does the page have 80 different stylesheets?

This website uses route based code splitting, it means that each route has its own js and css file that will be loaded once the user navigate to that page.
https://webpack.js.org/guides/code-splitting/

Here you see the 80 css files being downloaded upfront but they are only being prefetched with a low priority : 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ

I'm preparing a pull request to ignore those `<link>` nodes.

And an interesting article about preload/prefetch : https://medium.com/reloading/preload-prefetch-and-priorities-in-chrome-776165961bbf

peterbe:
I don't think `minimalcss` has a problem with prefetching in general. If you can, make a really simple HTML page with a simple single .css prefetch and check if `minimalcss` works with that.

peterbe:
> make a really simple HTML page with a simple single .css prefetch and check if minimalcss works with that.

I spoke too soon. You already made [one](https://github.com/peterbe/minimalcss/pull/275/files#diff-e3831af2fc4314f2b75c84a81f382497R310).

PR #275, opened 2018-10-18
Ignore preloaded and prefetched css

nicolas-t:
Following the discussion here : https://github.com/peterbe/minimalcss/issues/274

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -325,6 +325,8 @@ const processPage = ({
           elem.tagName === 'STYLE' &&
           (!elem.type || elem.type.toLowerCase() === 'text/css');
         const isStylesheetLink = elem =>
+          elem.rel !== 'preload' &&
+          elem.rel !== 'prefetch' &&
           elem.tagName === 'LINK' &&
           elem.href &&
           (elem.rel.toLowerCase() === 'stylesheet' ||
@@ -334,9 +336,6 @@ const processPage = ({
             elem.href.toLowerCase().startsWith('blob:') ||
             elem.media.toLowerCase() === 'print'
           );
-
-        const isResourceHinted = elem =>
-          elem.rel === 'preload' || elem.rel === 'prefetch';
 
         // #fragments are omitted from puppeteer's response.url(), so
         // we need to strip them from stylesheet links, otherwise the
@@ -353,9 +352,6 @@ const processPage = ({
         // of all the ones we're going to assess. For style elements,
         // also extract each tag's content.
         Array.from(document.querySelectorAll('link, style')).forEach(elem => {
-          if (isResourceHinted(elem)) {
-            return;
-          }
           if (isStylesheetLink(elem)) {
             const href = defragment(elem.href);
             hrefs.push(href);
--- a/src/run.js
+++ b/src/run.js
@@ -334,6 +334,11 @@ const processPage = ({
             elem.href.toLowerCase().startsWith('blob:') ||
             elem.media.toLowerCase() === 'print'
           );
+
+        const isResourceHinted = elem =>
+          (elem.rel === 'preload' ||
+          elem.rel === 'prefetch');
+
         // #fragments are omitted from puppeteer's response.url(), so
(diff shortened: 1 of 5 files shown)


src/run.js
```diff
@@ -349,6 +353,9 @@ const processPage = ({
         // of all the ones we're going to assess. For style elements,
         // also extract each tag's content.
         Array.from(document.querySelectorAll('link, style')).forEach(elem => {
+          if (isResourceHinted(elem)) {
```
stereobooster: you can update `isStylesheetLink` instead.

```js
const isStylesheetLink = elem =>
          elem.tagName === 'LINK' &&
          elem.href &&
          (elem.rel.toLowerCase() === 'stylesheet') &&
          !(
            elem.href.toLowerCase().startsWith('data:') ||
            elem.href.toLowerCase().startsWith('blob:') ||
            elem.media.toLowerCase() === 'print'
          );
```

nicolas-t: yup, done.

src/run.js
```diff
@@ -325,6 +325,8 @@ const processPage = ({
           elem.tagName === 'STYLE' &&
           (!elem.type || elem.type.toLowerCase() === 'text/css');
         const isStylesheetLink = elem =>
+          elem.rel !== 'preload' &&
```
stereobooster: `elem.rel !== 'preload' && elem.rel !== 'prefetch' && elem.rel.toLowerCase() === 'stylesheet` can we collapse this into `elem.rel.toLowerCase() === 'stylesheet'`? @peterbe do you remember why it was done this way?

peterbe: Actually, I don't see it like that. This massive if statement says that it needs to be either `<link rel="Stylesheet" href="whatever">` or `<link href="specifically.css">`

So if you have `<link rel="whatever" href="whatever.css">` it *would* be let through. 

I don't know why we allow `<link href="specifically.css">` but I guess it makes sense. If you don't specify the `rel` attribute out of "laziness" then it should work. 

I think the really important change would be to make it...
```javascript
elem.rel.toLowerCase() === 'stylesheet' || (
  !elem.rel && elem.href.toLowerCase().endsWith('.css')
)
```

I.e. the only time we allow it in is if the `href` attribute ends with `.css` if and only if the `rel` attribute is not set.

peterbe: Same with Google Chrome and Safari and Safari iOS. 

Basically, I'm inclined to revise this whole if statement to just be:
```javascript
const isStylesheetLink = elem =>
  elem.tagName === 'LINK' &&
  elem.href &&
  elem.rel.toLowerCase() === 'stylesheet' &&
  !(
      elem.href.toLowerCase().startsWith('data:') ||
      elem.href.toLowerCase().startsWith('blob:') ||
      elem.media.toLowerCase() === 'print'
  );
```

peterbe: In fact, we can simplify it even more:

```javascript
const isStylesheetLink = elem =>
  elem.tagName === 'LINK' &&
  elem.href &&
  elem.rel.toLowerCase() === 'stylesheet' &&
  !elem.href.toLowerCase().startsWith('data:') &&
  !elem.href.toLowerCase().startsWith('blob:') &&
  !elem.media.toLowerCase() === 'print';
```

What do you think @stereobooster ?

stereobooster: Yes, I thought of the same thing. I just wonder why we didn't do it initially. I guess we'll never know

nicolas-t: Hello,

I found the time to update my branch.

I used the function you suggested: 
```js
const isStylesheetLink = elem =>
  elem.tagName === 'LINK' &&
  elem.href &&
  elem.rel.toLowerCase() === 'stylesheet' &&
  !elem.href.toLowerCase().startsWith('data:') &&
  !elem.href.toLowerCase().startsWith('blob:') &&
  !elem.media.toLowerCase() === 'print';
```

modified like this : 
```diff
-  !elem.media.toLowerCase() === 'print';
+  elem.media.toLowerCase() !== 'print';
```

All tests are passing. \o/

PR #296, opened 2019-02-04
Optimize DOM selector lookups by pre-warming by selectors' parents

peterbe:
Part of #287

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -485,15 +485,29 @@ const minimalcss = async options => {
     );
   }
 
+  // This is the protection against ever looking up the same CSS selector
+  // more than once. We can confidently pre-populate it with a couple that
+  // we're confident about.
+  const decisionsCache = { '*': true, body: true, html: true, '': true };
+
   // Now, let's loop over ALL links and process their ASTs compared to
   // the DOMs.
-  const decisionsCache = {};
   const isSelectorMatchToAnyElement = selectorString => {
     // Here's the crucial part. Decide whether to keep the selector
     // Find at least 1 DOM that contains an object that matches
     // this selector string.
     return doms.some(dom => {
       try {
+        // console.log(
+        //   'SELECTOR:',
+        //   selectorString,
+        //   selectorString in decisionsCache
+        // );
+        if (selectorString in decisionsCache) {
+          throw new Error(
+            `Selector (${selectorString}) has already been added to the decisionsCache.`
+          );
+        }
         return dom(selectorString).length > 0;
       } catch (ex) {
         // Be conservative. If we can't understand the selector,
@@ -519,7 +533,6 @@ const minimalcss = async options => {
       return;
     }
     const ast = stylesheetAsts[href];
-
     csstree.walk(ast, {
       visit: 'Rule',
       enter: function(node, item, list) {
@@ -548,6 +561,42 @@ const minimalcss = async options => {
             const selectorString = utils.reduceCSSSelector(
               csstree.generate(node)
             );
+
+            // Before we begin, do a little warmup of the decision cache.
+            // From a given selector, e.g. `div.foo p.bar`, we can first look
+            // up if there's an point by first doing a lookup for `div.foo`
+            // because if that doesn't exist we *know*  we can igmore more
+            // "deeper" selectors like `div.foo p.bar` and `div.foo span a`.
+            const selectorParentSelectors = utils.selectorParentSelectors(
+              selectorString
+            );
+
+            // If "selectorString" was `.foo .bar span`, then
+            // this `selectorParentSelectors` array will be
+            // `['.foo', '.foo .bar']`.
+            // If `selectorString` was just `.foo`, then
+            // this `selectorParentSelectors` array will be `[]`.
+            let bother = true;
+            selectorParentSelectors.forEach(selectorParentString => {
+              if (bother) {
+                // Is it NOT in the decision cache?
+                if (selectorParentString in decisionsCache === false) {
+                  decisionsCache[
+                    selectorParentString
+                  ] = isSelectorMatchToAnyElement(selectorParentString);
+                }
+                // What was the outcome of that? And if the outcome was
+                // that it was NOT there, set the 'bother' to false which
+                // will popoulate the decision cache immediately.
+                if (!decisionsCache[selectorParentString]) {
+                  bother = false;
+                  decisionsCache[selectorString] = false;
+                }
+              } else {
+                decisionsCache[selectorParentString] = false;
+              }
+            });
+
             if (selectorString in decisionsCache === false) {
               decisionsCache[selectorString] = isSelectorMatchToAnyElement(
                 selectorString
@@ -612,9 +661,9 @@ const minimalcss = async options => {
   // it too.
   csso.compress(allCombinedAst, cssoOptions);
   postProcessOptimize(allCombinedAst);
-
+  const finalCss = csstree.generate(allCombinedAst);
   const returned = {
-    finalCss: csstree.generate(allCombinedAst),
+    finalCss,
     stylesheetContents
   };
   return Promise.resolve(returned);
--- a/src/run.js
+++ b/src/run.js
@@ -498,16 +498,6 @@ const minimalcss = async options => {
     // this selector string.
     return doms.some(dom => {
       try {
-        // console.log(
-        //   'SELECTOR:',
-        //   selectorString,
-        //   selectorString in decisionsCache
-        // );
-        if (selectorString in decisionsCache) {
-          throw new Error(
-            `Selector (${selectorString}) has already been added to the decisionsCache.`
-          );
-        }
         return dom(selectorString).length > 0;
       } catch (ex) {
         // Be conservative. If we can't understand the selector,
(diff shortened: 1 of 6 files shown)


peterbe:
I used this page to test things: https://www.peterbe.com/plog-original.html

It uses Semantic-UI which is horribly verbose! (Hint view source and look at the CSS)
Because this patch introduces some changes to how it prepopulates the decision cache with some dead-obvious selectors, the output is different compared to `master`. Diff:

```diff
183c183,185
< * {
---
> *,
> :after,
> :before {
196,197d197
<   height: 100%;
<   font-size: 14px;
203,206d202
< body,
< h1 {
<   padding: 0;
< }
208a205,206
>   margin: 0.67em 0;
>   line-height: 1.28571429em;
210a209
>   padding: 0;
213,214d211
<   margin: 0.67em 0;
<   line-height: 1.28571429em;
231c228,233
< body {
---
> ::-webkit-file-upload-button {
>   -webkit-appearance: button;
>   font: inherit;
> }
> body,
> html {
232a235,239
> }
> html {
>   font-size: 14px;
> }
> body {
233a241
>   padding: 0;
259a268,279
> ::-webkit-selection {
>   background-color: #cce2ff;
>   color: rgba(0, 0, 0, 0.87);
> }
> ::-moz-selection {
>   background-color: #cce2ff;
>   color: rgba(0, 0, 0, 0.87);
> }
> ::selection {
>   background-color: #cce2ff;
>   color: rgba(0, 0, 0, 0.87);
> }
```
(red means output with `master` and green output with this branch)

Here's the [master.css](https://gist.github.com/peterbe/8961ce7d78b4de44bbe7f5cc4dfdabc6) and [branch.css](https://gist.github.com/peterbe/6c638d84cdefb1bfda629dbc9505cd1a) outputs prettierfied.

I'm pretty sure this is correct fix. Things like Semantic-UI does like...

```css
*, :before, :after {
  -webkit-box-sizing:inherit;box-sizing:inherit
}
```
should not be deleted from the final CSS. (not that I understand the CSS itself)

peterbe:
In terms of speed... 

### Before:

```
▶ run-forever.sh time node bin/minimalcss.js https://www.peterbe.com/plog-original.html -o /dev/null
To stop this forever loop created a file called stopnow.
E.g: touch stopnow

Now going to run 'time node bin/minimalcss.js https://www.peterbe.com/plog-original.html -o /dev/null' forever (or until stopnow exists)


real	0m6.837s
user	0m3.733s
sys	0m0.357s

real	0m6.728s
user	0m4.254s
sys	0m0.273s

real	0m6.713s
user	0m4.225s
sys	0m0.295s

real	0m6.175s
user	0m3.726s
sys	0m0.272s

real	0m6.206s
user	0m3.785s
sys	0m0.263s

real	0m6.341s
user	0m3.906s
sys	0m0.258s

real	0m6.187s
user	0m3.735s
sys	0m0.265s

real	0m6.165s
user	0m3.730s
sys	0m0.251s
```

### After:
```
▶ run-forever.sh time node bin/minimalcss.js https://www.peterbe.com/plog-original.html -o /dev/null
To stop this forever loop created a file called stopnow.
E.g: touch stopnow

Now going to run 'time node bin/minimalcss.js https://www.peterbe.com/plog-original.html -o /dev/null' forever (or until stopnow exists)


real	0m4.941s
user	0m2.470s
sys	0m0.239s

real	0m5.020s
user	0m2.521s
sys	0m0.234s

real	0m4.984s
user	0m2.564s
sys	0m0.231s

real	0m5.302s
user	0m2.787s
sys	0m0.244s

real	0m4.888s
user	0m2.454s
sys	0m0.233s

real	0m5.210s
user	0m2.481s
sys	0m0.233s

real	0m5.161s
user	0m2.769s
sys	0m0.248s

real	0m5.282s
user	0m2.841s
sys	0m0.248s

real	0m5.013s
user	0m2.609s
sys	0m0.238s

real	0m5.068s
user	0m2.671s
sys	0m0.248s

real	0m5.113s
user	0m2.664s
sys	0m0.252s
```

That means, on *median* the total processing time goes from 6.3s to 5.1s.

peterbe:
@stereobooster This one is quite complex but I think it's valuable. Not expecting a deep code review but if you could take a look I'd appreciate it. Thanks in advance.

src/utils.js
```diff
@@ -1,3 +1,19 @@
+const csstree = require('css-tree');
+
+/**
+ * Simple single argument memoize function. Only works if the first and
+ * only argument is a hashable. E.g. a string.
+ */
+function memoize(fn) {
+  const cache = {};
```
stereobooster: Probably not an issue in this case, but we can use LRU cache here, like https://github.com/isaacs/node-lru-cache or https://github.com/lukeed/tmp-cache

src/run.js
```diff
@@ -485,15 +485,29 @@ const minimalcss = async options => {
     );
   }
 
+  // This is the protection against ever looking up the same CSS selector
+  // more than once. We can confidently pre-populate it with a couple that
+  // we're confident about.
+  const decisionsCache = { '*': true, body: true, html: true, '': true };
```
stereobooster: What empty string selector `''` stands for?

src/run.js
```diff
@@ -485,15 +485,29 @@ const minimalcss = async options => {
...
     // this selector string.
     return doms.some(dom => {
       try {
+        // console.log(
+        //   'SELECTOR:',
+        //   selectorString,
+        //   selectorString in decisionsCache
+        // );
+        if (selectorString in decisionsCache) {
+          throw new Error(
```
stereobooster: if you throw Error it will gather stacktrace along the way, if you want to use throw for control flow you may throw plain string (theoretically it is cheaper). But I would simply `return true` in this case instead of throw

src/run.js
```diff
@@ -548,6 +561,42 @@ const minimalcss = async options => {
             const selectorString = utils.reduceCSSSelector(
               csstree.generate(node)
             );
+
+            // Before we begin, do a little warmup of the decision cache.
+            // From a given selector, e.g. `div.foo p.bar`, we can first look
+            // up if there's an point by first doing a lookup for `div.foo`
+            // because if that doesn't exist we *know*  we can igmore more
```
stereobooster: ig**m**ore

src/run.js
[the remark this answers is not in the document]
peterbe: Doubt it's any difference compared to all other much more expensive things. 

By the way, 
one thing I looked at as a micro-optimization was if I compile the [regex here](https://github.com/peterbe/minimalcss/blob/7b2ee75c8a7446239ce969faac8bad20c28040ec/src/utils.js#L15) as a module-level variable once, then the `reduceCSSSelector` function wouldn't need to create it every time. 

To benchmark this, I bombarded the function will thousands of random CSS selectors. Measured the total time to run `reduceCSSSelector(x) for x in randomSelectors` before and after moving the regex out from inside the function. It took steadily about 39-41ms to make all those calls to `reduceCSSSelector` and 39-41ms when the regex only had to be created once.

Point is, it was a waste of time in the grand scheme of I/O that takes order of magnitude more time.

src/run.js
peterbe: I just removed it now. It was there to "check myself". In an earlier draft of the code, something wasn't right and lookups were made when they had already been made.

src/run.js
stereobooster: I guess there is an error where we convert AST to selectors, it should be `:after` instead of ``. Or at least we can add comment

peterbe: I think it comes from [`reduceCSSSelector`](https://github.com/peterbe/minimalcss/blob/7b2ee75c8a7446239ce969faac8bad20c28040ec/src/utils.js#L13). In terms of CSS selector and DOM inspecting the thing to look for is before the `:`. But if the whole CSS selector is `:after { ... }` then his is what happens.

tests/utils.test.js
```diff
@@ -31,3 +31,31 @@ test('Test removeSequentialSemis', () => {
   // multiple semicolon sequences
   expect(f('a;b;;c;;;d;;;;')).toEqual('a;b;c;d;');
 });
+
+test('Test parentSelectors', async () => {
+  const f = utils.parentSelectors;
```
stereobooster:
```suggestion
  const f = utils.getParentSelectors;
```

peterbe:
Thank you @stereobooster !

It still saddens me that it takes ~5 seconds to execute against these monster CSS files. The blame lies with Semantic-UI for being so overly verbose in its style.

peterbe:
By the way, for `https://www.peterbe.com/plog-original.html` prior to this branch there had to be `3,285` calls to [`dom(selectorString).length > 0`](https://github.com/peterbe/minimalcss/blob/e62652c977cfaec0b128409252e8ade913f4a669/src/run.js#L501). As of this patch there now needs only to be 1,563 calls. 

So 3,285 `cheerio` DOM lookups to 1,563 essentially reduced the total lookup time from ~6s to ~5s.

src/run.js
peterbe: Bummer! I forgot this before I merged.

peterbe: Done https://github.com/peterbe/minimalcss/commit/d16f2c15b27eae1dd5cf1b2f18b71899f483e656

PR #312, opened 2019-04-01
Added isJavaScriptEnabled option

VladislavTkachuk:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "minimalcss",
-  "version": "0.8.2",
+  "version": "0.9.0",
```
peterbe: Please don't change the version. It's my job :)
Also, it's automated with npm scripts when I make the release.

README.md
```diff
@@ -182,6 +182,7 @@ will be thrown by `minimalcss`. If you know it's safe to ignore errors (for exam
 third-party webpages), set this to `true`.
 * `styletags` - If set to `true`, on-page `<style>` tags are parsed along with external stylesheets. By default, only external stylesheets are parsed.
 * `enableServiceWorkers` - By default all Service Workers are disabled. This option enables them as is.
+* `isJavaScriptEnabled` - By default JavaScript is enabled. This option allows disable JavaScript.
```
peterbe: This name isn't right. Look at the option above for example. "isJavascriptEnabled" is *state* not a command. 

Also, this option needs to either explain how it coincides with the (admittedly poorly named) `withoutjavascript` option, or mention this option where `withoutjavascript` is documented.

src/run.js
```diff
@@ -193,6 +193,7 @@ const processPage = ({
       options.withoutjavascript === undefined
         ? true
         : !!options.withoutjavascript;
+	const isJavaScriptEnabled = (options.isJavaScriptEnabled === undefined) || !!options.isJavaScriptEnabled;
```
peterbe: I see that you used the same pattern as for `const withoutjavascript = ...`  but I can't at all remember why it was done like that. A bunch more idiomatic style would be:
```javascript
const isJavaScriptEnabled = !!options.isJavaScriptEnabled;
```

As a matter of fact, a bunch of these lines above can be simplified. E.g.
```javascript
const debug = !!options.debug;
```
...is way better because it guarantees that the variable is definitely a boolean. 

I.e.

```javascript
>>> var options = {debug: []}
>>> const debug = options.debug || false
>>> debug
[]
```

src/run.js
```diff
@@ -283,7 +284,9 @@ const processPage = ({
 
       let response;
 
-      if (withoutjavascript) {
+	  if (!isJavaScriptEnabled) {
+		page.setJavaScriptEnabled(false);
```
peterbe: Not only is this missing the `await ` but I overall don't understand this change. 

By default, `minimalcss` will run the page 1 time; with JavaScript enabled. 
If you pass `options.withoutjavascript = true` it will run the page 2 times; one time without javascript and one time with. 

So what does `isJavascriptEnabled` do?

src/run.js
VladislavTkachuk: Now when isJavaScriptEnabled option is undefined script must be enabled on page.
I can rename this to disableJavaScript and after that const disableJavaScript = !!options.disableJavaScript;

src/run.js
VladislavTkachuk: If you pass isJavascriptEnabled = false it will run the page 1 time without javascript, because in some cases styles inserts dynamically by js, but i don't need them

peterbe: But `minimalcss` already has that functionality.

VladislavTkachuk: How can i get this?

peterbe: Let's start over. What is your whole purpose of this? Do you want the minimalcss ONLY when javascript is completely disabled?

VladislavTkachuk: Yes

peterbe:
Seems Travis is failing because the file tests/main.test.js isn't properly formatted with Prettier. Can you look into that?

peterbe:
The tests are failing because of an integration test that is not your fault. I'll try to fix it and then perhaps you can rebase or merge in master.

peterbe:
Ok. Waiting for this to land. https://github.com/peterbe/minimalcss/pull/321

peterbe:
Can you please rebase this PR now?

peterbe:
Sorry, it needs to be rebased again :(
This time I'm much more confident it's going to work.

peterbe:
I am so sorry for missing this! I must have missed that you did one last master merge-in.

peterbe:
This is shipped in version 0.8.3 now.

PR #321, opened 2019-05-07
test without docker

peterbe:
Let's see what it takes to be able to install `puppeteer` in TravisCI "natively".


peterbe:
I *think* the reason we used Docker was because `puppeteer` couldn't be built in regular Travis. Seems to work now.

PR #331, opened 2019-09-19
minimalcss takes into account css from iframe

### LINKED ISSUE #206: minimalcss takes into account css from iframe
Need to investigate. Example https://minimalcss.app/?url=https%3A%2F%2Fstereobooster.github.io%2F

peterbe:
Fixes #206


peterbe:
I don't know if it's good enough to check that the domains are different. If you have a page like:
```html
<body>
  <iframe src="/page.html"></iframe>
</body>
```
this test would "fail". But how likely is that??

PR #344, opened 2019-12-26
Add whitelist option

AlexDubok:
There are cases when the same webpage may look different depending on user language, experiment state, etc. It would be great if you could pass selectors whitelist to tell minimalcss that these selectors should be left in final CSS, no matter if they are not present at the time of puppeteer rendering.

### DIFF
--- a/src/run.js
+++ b/src/run.js
@@ -409,13 +409,13 @@ const processPage = ({
  * @return Promise<{ finalCss: string, stylesheetContents: { [key: string]: string } }>
  */
 const minimalcss = async options => {
-  const { urls, whitelist = [] } = options;
+  const { urls } = options;
   const debug = options.debug || false;
   const cssoOptions = options.cssoOptions || {};
   const enableServiceWorkers = options.enableServiceWorkers || false;
   const puppeteerArgs = options.puppeteerArgs || [];
-  const isInWhiteList = selector =>
-    whitelist.some(rule => new RegExp(rule).test(selector));
+  const whitelist = options.whitelist || [];
+  const whitelistRules = whitelist.map(rule => new RegExp(rule));
 
   if (!enableServiceWorkers) {
     puppeteerArgs.push('--enable-features=NetworkService');
@@ -498,13 +498,12 @@ const minimalcss = async options => {
   // Now, let's loop over ALL links and process their ASTs compared to
   // the DOMs.
   const isSelectorMatchToAnyElement = selectorString => {
+    if (whitelistRules.some(regex => regex.test(selectorString))) {
+      return true;
+    }
     // Here's the crucial part. Decide whether to keep the selector
     // Find at least 1 DOM that contains an object that matches
     // this selector string.
-    if (isInWhiteList(selectorString)) {
-      return true;
-    }
-
     return doms.some(dom => {
       try {
         return dom(selectorString).length > 0;
--- a/src/run.js
+++ b/src/run.js
@@ -405,15 +405,18 @@ const processPage = ({
 
 /**
  *
- * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any, puppeteerArgs: Array<string>, cssoOptions: Object, ignoreCSSErrors?: boolean, ignoreJSErrors?: boolean, styletags?: boolean, enableServiceWorkers?: boolean, disableJavaScript?: boolean }} options
+ * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any, puppeteerArgs: Array<string>, cssoOptions: Object, ignoreCSSErrors?: boolean, ignoreJSErrors?: boolean, styletags?: boolean, enableServiceWorkers?: boolean, disableJavaScript?: boolean, whitelist?: Array<string> }} options
  * @return Promise<{ finalCss: string, stylesheetContents: { [key: string]: string } }>
  */
 const minimalcss = async options => {
-  const { urls } = options;
+  const { urls, whitelist = [] } = options;
   const debug = options.debug || false;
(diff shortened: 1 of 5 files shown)


src/run.js
```diff
@@ -405,15 +405,18 @@ const processPage = ({
 
 /**
  *
- * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any, puppeteerArgs: Array<string>, cssoOptions: Object, ignoreCSSErrors?: boolean, ignoreJSErrors?: boolean, styletags?: boolean, enableServiceWorkers?: boolean, disableJavaScript?: boolean }} options
+ * @param {{ urls: Array<string>, debug: boolean, loadimages: boolean, skippable: function, browser: any, userAgent: string, withoutjavascript: boolean, viewport: any, puppeteerArgs: Array<string>, cssoOptions: Object, ignoreCSSErrors?: boolean, ignoreJSErrors?: boolean, styletags?: boolean, enableServiceWorkers?: boolean, disableJavaScript?: boolean, whitelist?: Array<string> }} options
  * @return Promise<{ finalCss: string, stylesheetContents: { [key: string]: string } }>
  */
 const minimalcss = async options => {
-  const { urls } = options;
+  const { urls, whitelist = [] } = options;
```
peterbe: I know this works and it's compact but the pattern of the existing code is that all the optional options are on one line each. `urls` is the only one that isn't optional. 
If you can change this to (somewhere below):
```
const whitelist = options.whitelist || [];
```

src/run.js
```diff
@@ -405,15 +405,18 @@ const processPage = ({
...
   const debug = options.debug || false;
   const cssoOptions = options.cssoOptions || {};
   const enableServiceWorkers = options.enableServiceWorkers || false;
   const puppeteerArgs = options.puppeteerArgs || [];
+  const isInWhiteList = selector =>
+    whitelist.some(rule => new RegExp(rule).test(selector));
+
```
peterbe: I don't see the point of this being a standalone function. It's not curried either. And maybe that's the key here. 

If you have more than 0 whitelist selectors, and you're processing 1,000 selectors you have to execute `new RegExp(whitelist[i])` 1,000 times. And if you have 3 whitelist selectors you might be executing `new RegExp(whitelist[i])` 3,000 times. Or am I missing something?

A simple curry could be to make an array of regexes instead. Perhaps:
```javascript
const whitelistRules = whitelist.map(rule => new RegExp(rule));

... // somewhere around line 504

if (whitelistRules.some(regex => regex.test(selectorString))) {
  return true;
```

src/run.js
```diff
@@ -498,6 +501,10 @@ const minimalcss = async options => {
     // Here's the crucial part. Decide whether to keep the selector
     // Find at least 1 DOM that contains an object that matches
     // this selector string.
+    if (isInWhiteList(selectorString)) {
```
peterbe: The inclusion of these 4 lines now make it so that the code comment above (up to line 503) seems to belong to these 4 lines. Only. Can you re-arrange that a bit?

src/run.js
```diff
@@ -498,6 +501,10 @@ const minimalcss = async options => {
...
+      return true;
```
peterbe: Shouldn't this be added to `decisionCache`?
I.e. if you have `options.whitelist = ['.foo']` and something like:
```html
<p class="foo">One</p>
<p class="foo">Two</p>
```
it should only ever need to call `isInWhiteList` once. Right? 

It's been a while since I wrote the `decisionCache` implementation but the name is pretty clear :)
Can you take a look?

tests/examples/whitelist-css.css
```diff
@@ -0,0 +1,15 @@
...
+
+.icon {
+	height: 100%;
+}
+
+.icon-arrow {
+	width: 10px;
+}
+
+.icon-search {
```
peterbe: I don't see how this is different from the test of `.icon-arrow` but what could make this line interesting is...:
```suggestion
.icon-search, .not-icon-something {
```

peterbe:
One thing I've long been thinking about, which would make the whitelist feature you're adding "obsolete" is the ability to render the page in more non-trivial ways. 
The the moment, when `puppeteer` is asked to open the page and snaphot the DOM it simply opens the page. Nothing more. An integrator might want to consider more things and perhaps have a whole list of things that could happen. So perhaps by default, the `puppeteer` interactions should be like this:
```javascript
// the default opener
response = await page.goto(pageUrl, { waitUntil: 'networkidle0' });

// simulate open the page with `?advancedsearch=open`
response = await page.goto(pageUrl, { waitUntil: 'networkidle0' }, (dom, callback) => {
  dom.querySelector('.avanced-search').click();
  callback();
});

// simulate having the zh-CN locale which means different things will appear
response = await page.goto(pageUrl + '?locale=zh-CN', { waitUntil: 'networkidle0' });

// simulate opening with a huge viewport which means certain things appear differently
...
```

It might not be super trivial to implement but certainly powerful.

src/run.js
AlexDubok: Good point! Thank you!

src/run.js
AlexDubok: Sure, in the current implementation, the result of the whole function isSelectorMatchToAnyElement is added to the decisionCache. And since we are doing whitelist checks inside this function, the result is being added to the decisionCache too.

src/run.js
AlexDubok: Oh, now I see it :) Done.

AlexDubok:
Thank you for your feedback! I really appreciate that.
The "interactive" mode you're talking about is good and very interesting, but I think there still will be a place for a simple whitelist pattern.  
For example, in my case, the page could look differently directly after render depending on a number of A/B tests. 
Currently, the "different locale" problem could be solved by passing an array of URLs with different locales, but it would be much simpler if we had some way to preserve specific selectors rather then build such arrays for every route.

peterbe:
Version 0.9.0 released.

Issue #347, opened 2020-01-22
Get HTML retrieved by minimalcss 


gijo-varghese:
Is there any way to get the HTML retrieved while generating minimal css?

Right now I've to open the web page twice to get minimal css and extract html (using a custom script)

peterbe:
Inside the core of `minimalcss` it does take snapshots of the DOMs, which it then proceeds to work with to make the `finalCss`. I don't think it would be hard to add to the final returned output some of those DOMs as serialized strings. 
I can help you if you want to attempt to write a simple unit test.

PR #349, opened 2020-01-23
Get HTML retrieved by minimalcss

gijo-varghese:
Closes https://github.com/peterbe/minimalcss/issues/347


peterbe:
I wonder, why turning it into HTML? Why not leave it as an array of DOMs. Then, if someone decides they want to turn that into a serialized string, *they* can use the `doms[1].html()` for example. 
That way, you could, for example, check stuff on the DOM for presence.

gijo-varghese:
Done

PR #365, opened 2020-04-02
Added ignore request error option

Fgruntjes:
Well the name says it all, we needed a way to ignore request errors. In our case 404 errors.


Fgruntjes:
@peterbe thnx for the review. Have added a test and the docs line as you suggested.

peterbe:
Thanks @Fgruntjes

